PR #5, opened 2022-07-21
Use Jinja for filling code generation templates

pac48:


generate_parameter_library_py/generate_parameter_library_py/cpp_templates/validators.hpp
```diff
@@ -84,3 +84,18 @@ Result validate_int_array_bounds(const rclcpp::Parameter& parameter,
...
+  if (collection.find(string_param) == collection.end()) {
+    std::stringstream ss;
+    for (auto const& c : collection) ss << c << ", ";
+    return ERROR(
+        "The parameter (%s) with the value (%s) not in the valid values: [%s]",
+        parameter.get_name(), string_param, ss.str());
+  }
+
+  return OK;
+}
```
tylerjw, line 101: This could be a generic function so you could use this for sets of any type.

pac48: That is a good point. I went and made the change. Now that you bring it up, the other validation methods can likely also be templated to save on redundancy.

tylerjw (approved):

PR #20, opened 2022-07-27
Runtime parameters created from other parameters  

pac48:


generate_parameter_library_py/generate_parameter_library_py/jinja_templates/parameter_listener
```diff
@@ -58,6 +59,18 @@ rclcpp::Time __stamp;
...
+      // TODO remove any destroyed dynamic parameters
+{%- filter indent(width=6) %}
+{{remove_dynamic_parameters}}
+{%- endfilter %}
+      // declare any new dynamic parameters
+      rclcpp::Parameter param;
+{%- filter indent(width=6) %}
+{{update_declare_dynamic_parameters}}
+{%- endfilter %}
+    }
```
tylerjw, line 72: This function returns a bool so it needs a return statement at the end.

```
/home/tyler/code/ws_gen_ros2_control/build/generate_parameter_library_example/admittance_controller_parameters/include/admittance_controller_parameters.hpp:392:5: warning: no return statement in function returning non-void [-Wreturn-type]
  392 |     }
      |     ^
---
```

tylerjw (approved):
Tested and this works. Thank you for putting this together so quickly.

PR #32, opened 2022-07-29
parameter validators interface library

tylerjw:
Signed-off-by: Tyler Weaver <tylerjw@gmail.com>

This one requires more work to be integrated but I've created a library you can just include and use like this:

```c++
#include <parameter_validators/parameter_validators.hpp>

namespace pv = parameter_validators;

auto result = pv::contains<std::string>(param, {"a", "b"});
```


pac48 (approved):
LGTM

PR #44, opened 2022-07-31
Fixed length arrays

pac48:


tylerjw (approved):
This looks good. Have you tried making PRs to ros2_control to see what using this would be like?

PR #46, opened 2022-08-02
Fix scientific notation

pac48:


tylerjw (approved):
Tested and this works. I'll push a patch release.

PR #58, opened 2022-08-31
I think this is more 'correct' ROS CMakeLists.txt file

destogl:


README.md
```diff
@@ -65,9 +65,9 @@ generate_parameter_library(
 
 add_executable(minimal_node src/turtlesim.cpp)
 target_link_libraries(minimal_node PRIVATE
-  rclcpp::rclcpp
   turtlesim_parameters
 )
+ament_target_dependencies(minimal_node rclcpp)
```
tylerjw, line 70: I do not believe it is "more correct" to use ament_target_depenencies because it does not support INTERFACE/PUBLIC/PRIVATE linking and when you require the use of scoped linking using ament_target_dependencies prevents you from also using target_link_libraries. @ChrisThrasher

tylerjw: Is there some problem you are experiencing when using target_link_libraries that ament_target_dependencies solves?

ChrisThrasher: I'm not aware of any advantages `ament_target_dependencies` gives us in this context.

destogl: In this concrete case, no, but on the other PR yes when having multiple `target_link_libraries` and “PRIVATE” linking.

I find the other method just much easier to read.

PR #60, opened 2022-09-06
Example component node usage

### LINKED ISSUE #56: Examples showing proper use within a Component library
Are there any examples of using this within a component library in a public repo? 

The usage within code is straightforward in this case, but I am trying to figure out:
1.) what a good project structure would look like when creating a publicly consumable Component package, and 
2.) what steps need to be taken in CMake to properly expose dependencies for both static and runtime composition in this scenario

tylerjw:
Fixes #56 

This also fixes some deadlock conditions that were possible and updates the example readme.


example/include/generate_parameter_library_example/minimal_publisher.hpp
```diff
@@ -0,0 +1,52 @@
...
+
+class MinimalPublisher : public rclcpp::Node {
+ public:
+  MinimalPublisher(const rclcpp::NodeOptions& options = rclcpp::NodeOptions());
+
+ private:
+  void timer_callback();
+
+  rclcpp::TimerBase::SharedPtr timer_;
+  std::shared_ptr<admittance_controller::ParamListener> param_listener;
```
destogl, line 46:
```suggestion
  std::shared_ptr<admittance_controller::ParamListener> param_listener_;
```

example/src/minimal_publisher.cpp
```diff
@@ -26,61 +26,48 @@
 // ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 // POSSIBILITY OF SUCH DAMAGE.
 
-#include "admittance_controller_parameters.hpp"
-#include "rclcpp/rclcpp.hpp"
+#include <rclcpp/rclcpp.hpp>
+
+#include <admittance_controller_parameters.hpp>
+#include <generate_parameter_library_example/minimal_publisher.hpp>
```
destogl, line 32: Usually, first include is the .hpp file with the same name. This is a strategy what I am always using, because this file is the most relevant.

This would be the order we are usually using in ros2_control

```suggestion
#include <generate_parameter_library_example/minimal_publisher.hpp>

#include <admittance_controller_parameters.hpp>
#include <rclcpp/rclcpp.hpp>
```

This could be another way, nevertheless, I would always put the direct `.hpp` file first. This is what some linters also recommend.

```suggestion
#include <generate_parameter_library_example/minimal_publisher.hpp>

#include <rclcpp/rclcpp.hpp>

#include <admittance_controller_parameters.hpp>
```

example/src/minimal_publisher.cpp
```diff
@@ -26,61 +26,48 @@
...
-      }
-    }
-  }
-
-  rclcpp::TimerBase::SharedPtr timer_;
-  std::shared_ptr<admittance_controller::ParamListener> param_listener;
-  admittance_controller::Params params_;
-};
-
-int main(int numArgs, const char** args) {
```
destogl, line 78: I understand that for the Compoennts one doesn't need `main`. Nevertheless, would it make sense to have both examples, Component Node and "normal" Node?

destogl (changes requested):
A few things go clarify before merging. Next time would be great to get such PR split into multiple ones.

We now have:

1. Fixing potential deadlocks in generated code
2. Renaming executable name in the CMakeLists.txt
3. Creating Component Node example.

example/src/minimal_publisher.cpp
tylerjw: When you create a component node it also creates an executable. You can see from the cmake and the readme of the example that there is still an executable and this file is now redundant.

example/src/minimal_publisher.cpp
tylerjw: These changes are now part of the latest version of this PR. The order of the includes is set by clang-format.

tylerjw:
@destogl These are now three separate PRs. Hopefully that makes this easier to review.

example/CMakeLists.txt
```diff
@@ -2,23 +2,40 @@ cmake_minimum_required(VERSION 3.22)
...
+)
+target_link_libraries(minimal_publisher
+  PUBLIC
+    admittance_controller_parameters
+    rclcpp::rclcpp
+    rclcpp_components::component
+)
+rclcpp_components_register_node(minimal_publisher
+  PLUGIN "admittance_controller::MinimalPublisher"
+  EXECUTABLE test_node
```
tylerjw, line 29: @destogl when you create a component node, it now creates a stand-alone executable to use that node with this macro. This is what makes the main.cpp redundant as it is now generated for us.

example/include/generate_parameter_library_example/minimal_publisher.hpp
tylerjw: Applied in the latest version.

example/src/minimal_publisher.cpp
tylerjw: Done

destogl (commented):
This looks fine to me. I am just not sure how to test it. I have tried using `ros2 run` but I am missing the parameters file to add as argument.

tylerjw:
The readme in the example directory has instructions for running it.

PR #63, opened 2022-09-06
Update README for example

tylerjw:
The readme for the exmaple is out of date. This updates it to the latest state of output from the various commands.


destogl (approved):
Great!

PR #64, opened 2022-09-06
Fix deadlock in update_dynamic_parameters

tylerjw:
This fixes a deadlock condition in update_dynamic_parameters. The reason for the deadlock is that when it was declaring new parameters, the mutex was locked and the update callback would be called that would also try to lock the mutex.

To fix this, I updated all the usage of the internal state `params_` to only be accessed through the getter method and a new private setter method. This means that the methods inside ParamListener only lock the mutex to copy or write to this internal state value instead of locking for the full duration of some of the methods that can cause callbacks to fire that might also need to lock the mutex.


destogl (approved):
Thanks! It looks good to me!

PR #66, opened 2022-09-08
Test validators and fix bugs

### LINKED ISSUE #59: Potentially misleading error messages in the library
When working to add a new controller that uses this library, I have found few misleading issues.

The code used is under: [destogl/tesko_controller](https://github.com/destogl/tesko_controller) repository. Download it and compile it.

When running tests (without YAML file) some parameters are not set, and it is expected for tests to fail. Nevertheless, the error messages in the test results are misleading.

For example, not having any joint in `joints` parameters (array<string>) set results in a message
```
Exception thrown during controller's init stage with message: parameter 'joints' has invalid type: expected [string] got [string_array]
```
And the error happens in line 172-175 or file `testko_controller_parameters.hpp` (generated file) where it is checked if string array is empty.

Commenting out those lines, the same issue (but opposite output) happens when `interface_name` parameter (string) is empty, i.e.:
```
Exception thrown during controller's init stage with message: parameter 'interface_name' has invalid type: expected [string_array] got [string]
```

Looking at the code, I can not really understand why is this happening.

Hopefully, with this minimal example, you will have an idea about that.

tylerjw:
Fixes #59 

This PR clarifies which validators can be used for which types of parameters in the README, fixes the odd bugs that @destogl found with vectors of strings, and adds unit tests for all the validator functions.

### DIFF
--- a/parameter_traits/test/validators_tests.cpp
+++ b/parameter_traits/test/validators_tests.cpp
@@ -1,0 +1,460 @@
+// Copyright (c) 2022, PickNik Inc.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+//    * Redistributions of source code must retain the above copyright
+//      notice, this list of conditions and the following disclaimer.
+//
+//    * Redistributions in binary form must reproduce the above copyright
+//      notice, this list of conditions and the following disclaimer in the
+//      documentation and/or other materials provided with the distribution.
+//
+//    * Neither the name of the copyright holder nor the names of its
+//      contributors may be used to endorse or promote products derived from
+//      this software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+// ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+// LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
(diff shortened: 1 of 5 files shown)


parameter_traits/test/validators_tests.cpp
```diff
@@ -0,0 +1,460 @@
...
+  EXPECT_FALSE(bounds<double>(Parameter{"", -4.3}, 1.0, 5.0).success());
+  EXPECT_FALSE(bounds<double>(Parameter{"", 10.2}, 1.0, 5.0).success());
+
+  EXPECT_TRUE(bounds<int64_t>(Parameter{"", 1}, 1, 5).success());
+  EXPECT_TRUE(bounds<int64_t>(Parameter{"", 4}, 1, 5).success());
+  EXPECT_TRUE(bounds<int64_t>(Parameter{"", 5}, 1, 5).success());
+  EXPECT_FALSE(bounds<int64_t>(Parameter{"", -4}, 1, 5).success());
+  EXPECT_FALSE(bounds<int64_t>(Parameter{"", 10}, 1, 5).success());
+
+  EXPECT_TRUE(bounds<bool>(Parameter{"", true}, false, true).success());
```
Abishalini, line 55: I'm a little confused why you would use bounds with bool type?

parameter_traits/test/validators_tests.cpp
```diff
@@ -0,0 +1,460 @@
...
+  EXPECT_FALSE(bounds<bool>(Parameter{"", true}, false, false).success());
+
+  EXPECT_ANY_THROW(bounds<int64_t>(Parameter{"", ""}, 1, 5));
+  EXPECT_ANY_THROW(bounds<std::string>(Parameter{"", 4}, "", "foo"));
+}
+
+TEST(ValidatorsTests, LowerBounds) {
+  EXPECT_TRUE(lower_bounds<double>(Parameter{"", 2.0}, 2.0).success());
+  EXPECT_TRUE(lower_bounds<double>(Parameter{"", 4.3}, 1.0).success());
+  EXPECT_FALSE(lower_bounds<double>(Parameter{"", -4.3}, 1.0).success());
```
Abishalini, line 66: Why call it lower_bounds instead of something like greater_than_or_equal?

parameter_traits/test/validators_tests.cpp
tylerjw: I don't expect someone to actually use these with boolean types, I just wanted to test that they work as expected if you do.

parameter_traits/test/validators_tests.cpp
tylerjw: I started with a single `bounds` function and these grew out of that. When working through this yesterday I had a similar thought about making these map better for operators.

destogl (approved):
Thanks! This solves the issue!!

destogl (approved):

PR #68, opened 2022-09-15
🈵 Support use of '_' in mapped parameters.

destogl:
Adding support to map parameters which have "_" in their name.
This also make the code more about explicit keywords and without assumption about name format.

Check the extended example parameter setup for more details.


tylerjw (approved):
This seems like a good change. I expect you've tested it.

PR #91, opened 2022-11-12
Migrate from parameter_traits to RSL (take 2)

tylerjw:


tylerjw:
```error: no matching function for call to ‘rclcpp::ParameterValue::ParameterValue(rsl::StaticString<25>&)’```

- [ ] convert static strings and vectors into parameter value types using function

ChrisThrasher (approved):

PR #92, opened 2022-11-16
Make it easy for users to override

tylerjw:
https://colcon.readthedocs.io/en/released/user/overriding-packages.html#how-to-make-it-easier-for-your-users-to-override

### DIFF
--- a/example/CMakeLists.txt
+++ b/example/CMakeLists.txt
@@ -15,8 +15,8 @@ add_library(minimal_publisher SHARED
   src/minimal_publisher.cpp
 )
 target_include_directories(minimal_publisher PUBLIC
-  $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
-  $<INSTALL_INTERFACE:include>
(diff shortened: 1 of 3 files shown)


example/CMakeLists.txt
```diff
@@ -15,8 +15,8 @@ add_library(minimal_publisher SHARED
   src/minimal_publisher.cpp
 )
 target_include_directories(minimal_publisher PUBLIC
-  $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
-  $<INSTALL_INTERFACE:include>
+  $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>
+  $<INSTALL_INTERFACE:include/generate_parameter_library_example>
```
ChrisThrasher, line 19: Does this example actually get installed or is this just an example of how someone would install a project?

tylerjw: It is just an example project but I would like it to be as correct as possible in case someone copies from it.

ChrisThrasher (approved):

PR #93, opened 2022-11-17
Add keyword INTERFACE to fix build error 'ar: no archive members specified' on macOS

light-tech:
On macOS, CMake (3.23) could not build the generated target: It cannot produce the static library with the error
```
ar: no archive members specified
```
because there is no object file supplied (the generated target is header-only). Simply adding the keyword `INTERFACE` disables building the static library.

### DIFF
--- a/generate_parameter_library/cmake/generate_parameter_library.cmake
+++ b/generate_parameter_library/cmake/generate_parameter_library.cmake
@@ -74,13 +74,13 @@ function(generate_parameter_library LIB_NAME YAML_FILE)
   )
 
   # Create the library target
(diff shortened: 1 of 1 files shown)


tylerjw (approved):

tylerjw:
I'm not sure why I did not do it this way in the first place. Approving so it'll run CI.

light-tech:
> I'm not sure why I did not do it this way in the first place. Approving so it'll run CI.

I believe some `ar` implementation allows for an empty list of object files, in which case, it simply produces an empty `.a` file (reportedly just 8 bytes in size).

Issue #96, opened 2022-12-12
Specify range constraint in parameter descriptor when bounds<> validator is present


bjsowa:
The ParameterDescriptor message contains fields for specifying the range for float and int parameters: 

https://github.com/ros2/rcl_interfaces/blob/43ffb7b6df6a98a1a35d0932d148ad12de549594/rcl_interfaces/msg/ParameterDescriptor.msg#L27-L36

This is used, for example, in [rqt_reconfigure](https://github.com/ros-visualization/rqt_reconfigure/tree/rolling) to display sliders for choosing the parameter value.

I would love it if these fields were actually used in this library,. Seems like they can be easily populated from the values passed to the bounds<> validator.

PR #103, opened 2023-04-11
Populate Range Constraints in Parameter Descriptors from Validation Functions

chancecardona:
This PR resolves [Issue 96](https://github.com/PickNikRobotics/generate_parameter_library/issues/96). 

This uses the `declare_parameter` and `declare_runtime_parameter` templates to populate the parameter descriptor. 
For this, some string checking on the validation function was added to see if it was appropriate to populate the Descriptor range constraints.  In the jinja templates I simply check against the string name of the validation function and the number of arguments. This should work in all cases here, but strange things may happen if another validation function with a substring of "gt" or "lt" with 1 argument, or "bounds" with 2 args is either created in RSL or in a user defined validation function. For this purpose I figure this is acceptable. 

This also may not allow setting of the descriptors dynamically (only in the declaration) but that can be a future work (unless it's a simple change) if it's a desired feature.

This should work for the RSL validation functions: `bounds`, `element_bounds`, `lower_bounds`, `upper_bounds`, `gt_eq`, `lt_eq`, `gt`, and `lt` validation functions by using std::numerical_limits library to populate the corresponding `to_value` or `from_value` that's not defined with either `max()` or `lowest()` for that parameter cpp type.

To assist in Unit Tests, a `get_descriptors` method was also added in the `parameter_library_header` template.
Finally, a Unit Test was created in the `example` directory and should run when `colcon test` is run.

### DIFF
--- a/generate_parameter_library_py/generate_parameter_library_py/jinja_templates/parameter_library_header
+++ b/generate_parameter_library_py/generate_parameter_library_py/jinja_templates/parameter_library_header
@@ -5,6 +5,7 @@
 #include <algorithm>
 #include <array>
 #include <functional>
+#include <limits>
 #include <mutex>
 #include <rclcpp/node.hpp>
 #include <rclcpp_lifecycle/lifecycle_node.hpp>
--- a/generate_parameter_library_py/generate_parameter_library_py/jinja_templates/parameter_library_header
+++ b/generate_parameter_library_py/generate_parameter_library_py/jinja_templates/parameter_library_header
@@ -113,6 +113,11 @@ struct StackParams {
       return params_;
     }
 
+    std::vector<rcl_interfaces::msg::ParameterDescriptor> get_descriptors(const std::vector<std::string> & names) const{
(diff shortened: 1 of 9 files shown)


generate_parameter_library_py/generate_parameter_library_py/jinja_templates/parameter_library_header
```diff
@@ -113,6 +114,11 @@ struct StackParams {
       return params_;
     }
 
+    std::vector<rcl_interfaces::msg::ParameterDescriptor> get_descriptors(const std::vector<std::string> & names) const{
+      std::lock_guard<std::mutex> lock(mutex_);
+      return parameters_interface_->describe_parameters(names);
```
tylerjw, line 119: I'm not sure we need this method. If it just calls `describe_parameters` on the node we can also just call that in the tests.

tylerjw (commented):
One small comment, otherwise, this looks good. Thank you for this nice feature :champagne:

PR #105, opened 2023-04-13
Fix Parameter Descriptor Incorrectly Populating Range Constraints for size_lt and size_gt

chancecardona:
Due to the string checking in the declare(_runtime)_parameters templates, `size_lt` and `size_gt` would incorrectly populate the descriptor range constraints since it also has 1 argument. This changes that line in the template to explicitly check for `lt`, `lt_eq`, `gt`, and `gt_eq` so that we avoid this error and hopefully other errors in the future. 

Not the cleanest implementation, but added a function_base_name without the template brackets `<>`. This passes the checks I've looked for and should reduce the descriptors being inadvertently filled unless the validation function name has both the substring `lesser`/`greater` and `bounds` in it with 1 argument. Could clean up in future if desired but would need to think about how to structure this in the jinja templates / main.py, as it's a bit long. But this should work without any false positives now.


tylerjw (approved):
Thank you for catching this. How useful do you think it would be for me to push out a patch release for this fix right away?

PR #106, opened 2023-04-19
Log when a parameter is set

### LINKED ISSUE #86: Option to print parameter values
I have been using this library and I am really happy with it - thanks for the contribution!

This issue is more of a question/feature request: can we log the parameters that are loaded by the parameter listener? This would be very useful for debugging purposes.

tylerjw:
Fixes #86

If you build this locally, you can test this by running the example node with debug logging:
```shell
ros2 run generate_parameter_library_example test_node --ros-args --params-file src/generate_parameter_library/example/config/implementation.yaml --log-level debug
```

Then in another terminal change a parameter:
```shell
ros2 param set /admittance_controller control.frame.id new_frame
```

You should see an output like this in the logs:
```
[DEBUG] [1681944620.098785049] [admittance_controller]: control.frame.id: string = new_frame
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -421,3 +421,20 @@ class ParamListener {
 } // namespace cpp_namespace
 ```
 The structure of the `Params` struct and the logic for declaring and updating parameters is generated from a YAML file specification.
+
+# FAQ
+
+Q. What happens if I declare a parameter twice? Will I get an error at runtime?
+A. The declare routine that is generated checks to see if each parameter has been declared first before declaring it. Because of this you can declare a parameter twice but it will only have the properties of the first time you declared it. Here is some example generated code.
+```cpp
+if (!parameters_interface_->has_parameter(prefix_ + "scientific_notation_num")) {
+    rcl_interfaces::msg::ParameterDescriptor descriptor;
+    descriptor.description = "Test scientific notation";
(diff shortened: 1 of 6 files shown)


abake48 (approved):
tested and lgtm

tylerjw:
Thank you

christophfroehlich (commented):
It works, LGTM (except for the missing dash in your instructions ;) `--params-file`).

I'm just wondering: At the first run, all parameters are listed in the debug output. After changing .frame.id, there is an output line with the change of this parameter but also a subset of other parameters (pid. and gains. structure. I guess all __map - parameters?). Is this intentional?

christophfroehlich:
ok, I just saw that at the beginning "all" parameters are written, at run-time only the changed one and the __map parameters.

christophfroehlich (approved):

tylerjw:
> ok, I just saw that at the beginning "all" parameters are written, at run-time only the changed one and the __map parameters.

The reason the map parameters are re-read after any parameter change is it accounts for if you change the key parameter for the map. Re-reading in the library also will declare any parameters that have not been declared and give them default values in the map. Eventually, I hope to fix this by detecting when the key is changed and only re-reading the map parameters then.

Issue #98, opened 2023-03-08
Python Parameter Interface


Flova:
Are there any plans regarding a Python integration?

The current handling of parameters in Python under stock ROS2 suffers from similar issues compared to C++.
While Python is dynamic enough to handle generation of a config class during runtime, code generation during build time could improve the experience while autocompleting parameter names in the IDE etc..

I assume integration of CMakeLists.txt based (python) packages is easier compared to setup.py based ones.

tylerjw:
I don't regularly use the python interface to ROS so I had no idea it also needed something like this but if you wanted to work on a python code generation output I'd be happy to try to review it and merge your work. We did use jinja templates and python to generate so maybe that would also work for python?

tylerjw:
Python interface was just merged. Check it out!

PR #110, opened 2023-05-15
Add Python support for generate_parameter_library

pac48:
This PR adds Python support for generate_parameter_library, which is a feature that has already been requested https://github.com/PickNikRobotics/generate_parameter_library/issues/98. All existing features should be supported for the generated Python module: including:
1. integration into the build system (e.g. setup.py)
1. automatically declaring and updating parameters 
1. automatically declaring and updating runtime parameters 
1. parameter validation
1. supported validators
1. handle user-defined validators

I added an example for the generated Python module in [minimal_publisher.py](https://github.com/pac48/generate_parameter_library/blob/add-generate-parameter-library-python-module/examples/python/generate_parameter_module_example/minimal_publisher.py). 

Note that the Python code does not separate out stack and heap-allocated parameters and is not designed for multi-threading.


tylerjw (approved):
:tada: thank you for this! Looking forward to the docs generation PR too.

PR #127, opened 2023-06-28
Fixed incorrect usage of fmt, causing compile errors

sprenger120:
Just found the library and I really like it. It saves so much time; thank you for your work. There is just a tiny error in the code generator, causing the following error

```
error: temporary of non-literal type ‘std::__cxx11::basic_string<char>’ in a constant expression
  256 |           throw rclcpp::exceptions::InvalidParameterValueException(fmt::format("Invalid value set during initialization for parameter 'led_strip.block_length': " + validation_result.error()));
``` 
The fix is easy, replace the "+ validation_result" with ", validation_result" and add a "{}" for placeholding. That is all this PR does

### DIFF
--- a/generate_parameter_library_py/generate_parameter_library_py/cpp_convertions.py
+++ b/generate_parameter_library_py/generate_parameter_library_py/cpp_convertions.py
@@ -92,7 +92,7 @@ class CPPConverstions:
         return (
             f"throw rclcpp::exceptions::InvalidParameterValueException"
             f'(fmt::format("Invalid value set during initialization for '
-            f"parameter '{param_name}': \" + validation_result.error()));"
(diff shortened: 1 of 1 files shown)


sjahr (approved):
I don't have issues building this package without this change but building it with this works too :+1: @pac48 Can you take a look at this?

sprenger120:
@sjahr Did you try forcing C++20? I forgot to mention that

tylerjw (approved):

tylerjw:
Thank you for this fix! I'm running CI to ensure it doesn't break anything we already tested for, but the change looks good. I will add an issue to add a test to force C++20 in one of the CI jobs so we can catch this in the future. I don't want to make it hard for people to use C++20 by accident.

PR #133, opened 2023-07-24
Replace allocator with validator

fmauch:
The example for adding a custom validator function was labeled "Here is an example custom allocator."

I assume this is kind of a false friend, as the devs more often are confronted with custom allocators rather than validators ;-)


tylerjw:
I'm confused; this is documenting a custom validator, not a custom allocator.

tylerjw:
You were right, we somehow have the wrong word there. Re-opening and merging this.

PR #134, opened 2023-07-24
Add documentation about generating a Markdown documentation

fmauch:
I was happily surprised that that feature has been merged, but it seems not to be documented. This adds a short section to the main README that explains how to generate a markdown documentation.

On a side note: I was quite surprised that the script to generate the docs is in the global PATH. I was expecting a `ros2 run generate_parameter_library_py generate_parameter_library_markdown`, but that might just be me having false expectations combined with this not being documented.


tylerjw:
It isn't documented yet because it is still missing some features. It is more of a prof-of-concept than a full-featured solution to auto-generating docs. Thank you for the PR; I'll happily merge this, though. If you use this, any improvements you want to make and submit PRs for would be welcome.

PR #146, opened 2023-09-11
Remove `my_parameter` from python template

jaagut:
I have tested your great library for generating parameters for my Python node and noticed that the parameter `my_parameter` was always included. I presume this was accidental.

This PR removes the `my_parameter` from the `parameter_library_header` template.


pac48 (approved):
Thank you for your contribution. You are correct that the addition of that parameter is a mistake.

tylerjw (approved):

Issue #158, opened 2023-12-07
Feature Request: Use generate_parameter_module through ament_cmake_python


shu-rt:
Hi, thank you for the usefull library.  

I am currently working on a ROS 2 package that uses both C++ and Python codes.  
I have a node written in Python, so I want to use the `generate_parameter_module` instead of `generate_parameter_library`.  
Is there a way to use it through `ament_cmake_python`?

PR #161, opened 2023-12-17
Enable generate_parameter_module through ament_cmake_python

pac48:
This PR adds the functionality to generate a Python parameter module in a Cmake project. See [here](https://github.com/PickNikRobotics/generate_parameter_library/issues/158 ) for the feature request. I also added another Python example that uses this new feature.


example_cmake_python/README.md
```diff
@@ -0,0 +1,157 @@
+# Example:
+
+## Build the node
+
+```
+ mkdir colcon_ws
+ mkdir colcon_ws/src
```
ChrisThrasher, line 7:
```suggestion
 mkdir -p colcon_ws/src
```

generate_parameter_library/cmake/generate_parameter_library.cmake
```diff
@@ -92,6 +92,52 @@ function(generate_parameter_library LIB_NAME YAML_FILE)
...
+  set(LIB_INCLUDE_DIR ${CMAKE_CURRENT_BINARY_DIR}/${LIB_NAME})
+  file(MAKE_DIRECTORY ${LIB_INCLUDE_DIR})
+
+  find_package(PythonInterp REQUIRED)
+  execute_process(
+          COMMAND "${PYTHON_EXECUTABLE}" -c "import sys;v = sys.version.split()[0];v = v.split('.');print(f'python{v[0]}.{v[1]}')"
+          OUTPUT_VARIABLE PYTHON_VERSION
+          OUTPUT_STRIP_TRAILING_WHITESPACE)
+
+  set(PARAM_HEADER_FILE ${CMAKE_INSTALL_PREFIX}/local/lib/${PYTHON_VERSION}/dist-packages/${PROJECT_NAME}/${LIB_NAME}.py)
```
ChrisThrasher, line 119: This CMake function creates files in `/usr/local/lib`? Why do we want to do that?

I also don't think `${CMAKE_INSTALL_PREFIX}/local/lib` is the right path if we do know for a fact that this is what we want to do.

pac48: What we want to do is install the Python package to a place on the `PYTHONPATH`. The way ROS 2 goes about this is to add an entry to the `PYTHONPATH` for every Python package built from source in your workspace. In this context,  `CMAKE_INSTALL_PREFIX` refers to `ros_ws/install/my_python_pkg`.

tylerjw:
This is failing pre-commit. You might need to rebase it, it seems to be about python quotes.

tylerjw (approved):

Issue #151, opened 2023-10-03
Error with nested maps


g-argyropoulos:
While trying out the map feature of `generate_parameter_library`, I encountered a compilation error with nested maps. A minimal example is the following:

``` yaml
queries_list: {
    type: string_array,
  }

  queries:
    __map_queries_list:    
      arguments: {
        type: string_array,
      }
      __map_arguments:       
        command: {
          type: string,
        }    
```
The output looks like that:

```shell-session
In member function ‘void foo::ParamListener::refresh_dynamic_parameters()’:
error: ‘struct foo::Params’ has no member named ‘arguments’
  153 |       for (const auto & value : updated_params.arguments){
      |                                                ^~~~~~~~~
error: ‘struct foo::Params::Queries’ has no member named ‘__map_queries_list’; did you mean ‘MapQueriesList’?
  154 |           auto& entry = updated_params.queries.__map_queries_list.arguments_map[value];
      |                                                ^~~~~~~~~~~~~~~~~~
      |                                                MapQueriesList
In member function ‘rcl_interfaces::msg::SetParametersResult foo::ParamListener::update(const std::vector<rclcpp::Parameter>&)’:
error: ‘struct foo::Params’ has no member named ‘arguments’
  187 |         for (const auto & value : updated_params.arguments){
      |                                                  ^~~~~~~~~
error: ‘struct foo::Params::Queries’ has no member named ‘__map_queries_list’; did you mean ‘MapQueriesList’?
  190 |                 updated_params.queries.__map_queries_list.arguments_map[value].command = param.as_string();
      |                                        ^~~~~~~~~~~~~~~~~~
      |                                        MapQueriesList
In member function ‘void foo::ParamListener::declare_params()’:
error: ‘struct foo::Params’ has no member named ‘arguments’
  231 |       for (const auto & value : updated_params.arguments){
      |                                                ^~~~~~~~~
error: ‘struct foo::Params::Queries’ has no member named ‘__map_queries_list’; did you mean ‘MapQueriesList’?
  232 |           auto& entry = updated_params.queries.__map_queries_list.arguments_map[value];
      |                                                ^~~~~~~~~~~~~~~~~~
      |                                                MapQueriesList
```

Are nested maps not supported or am I doing something wrong?

tylerjw:
I don't have any examples of anyone else doing this. It is probably just not yet working. You are welcome to submit PRs if you want to work on making nested maps work.

PR #166, opened 2024-01-01
Support nested mapped parameters

pac48:
This issue https://github.com/PickNikRobotics/generate_parameter_library/issues/151 describes a problem when using multiple levels of nested maps. Currently, only one level of nesting is supported.


christophfroehlich (approved):
Hard to review the jinja temlates, but the code generated from the provided yaml looks good. 
And I tested it with ros2_controllers, it doesn't break current configs.

Should one update the README describing this?

sea-bass (approved):

christophfroehlich (approved):
Thanks!

PR #169, opened 2024-01-13
Change default max of floating point to infinity from max

griswaldbrooks:
# Description
I recently found while using the library that the default maximum value for a floating point value (for example when using `gt_eq`) was set to `std::numeric_limits<double>::max()` which in C++ is less than `std::numerical_limits<double>::infinity()`. In my use case infinity was the default value of the parameter and this caused problems when trying to use something like `rqt_reconfigure` to look at the parameter and caused the rest of the parameters not to load.

It makes sense to me that the largest positive value would be infinity and not max double.

# Testing
I updated and added a test in `descriptor_test_gtest.cpp` and ran
```
colcon test
```
but would like help adding additional tests that shows the default value of infinity.

I would also like to know if there's a way for my to run this test node so I can check the `rqt_reconfigure` output.
When I run
```
ros2 run generate_parameter_library_example test_node 
```
I get
```
terminate called after throwing an instance of 'rclcpp::exceptions::ParameterUninitializedException'
  what():  parameter 'fixed_string_no_default' is not initialized
[ros2run]: Aborted
```
:shrug:


pac48:
Thanks for fixing this.

> I would also like to know if there's a way for my to run this test node so I can check the rqt_reconfigure output.
When I run

To fix that, you just have to pass the parameter yaml into the node like this. 

```
ros2 run generate_parameter_library_example test_node --ros-args --params-file src/generate_parameter_library/example/config/implementation.yaml
```
Note, that the path is relative to the workspace.

pac48:
@griswaldbrooks  Do you mind running pre-commit to fix the formatting?

griswaldbrooks:
@pac48 have you ever seen something like
```shell
[WARN] [1705172458.735387611] [rqt_reconfigure]: Failed to retrieve parameters from node /admittance_controller: cannot convert float NaN to integer
[WARN] [1705172459.250283042] [rqt_reconfigure]: Failed to retrieve parameters from node: cannot convert float NaN to integer
```

this is very similar to what I encountered on my other project when attempting to use infinity

griswaldbrooks:
also, for more context, this happens when running `ros2 run rqt_reconfigure rqt_reconfigure` with my changes. When running
```shell
griswald@gpl-container-dev:~/ws$ ros2 param get admittance_controller default_infinity                    
Double value is: inf 
```
There doesn't seem to be an issue, but this affects running `rqt_reconfigure` and custom tools that are getting the parameter ranges.

christophfroehlich:
changed again by #378

Issue #170, opened 2024-01-16
Code generation gives syntax errors with newlines


wouterio:
A description containing newlines
```
  hover_override: {
    type: int,
    default_value: 1,
    description: "Override hover action:\n0: Hover\n1: Push\n2: Pull\n-1: Do not override",
    validation: {
      one_of<>: [ [ 0, 1, 2, -1 ] ],
    },
  }
```
leads to generated python code with syntax errors:
```
            if not self.node_.has_parameter(self.prefix_ + "hover_override"):
                descriptor = ParameterDescriptor(description="Override hover action:   <<< missing quote
                0: Hover
                1: Push
                2: Pull
                -1: Do not override", read_only = False)
                parameter = updated_params.hover_override
                self.node_.declare_parameter(self.prefix_ + "hover_override", parameter, descriptor)
```

christophfroehlich:
I had a similar issue with generated c++ code and implemented a jinja filter for this, see #164
Maybe you can use it (or implement something similar) for the python templates.

wouterio:
I could easily avoid the problem skipping the newlines. But I'd say it's still something to be fixed.

wouterio:
Ha that was quick, thank you!

PR #176, opened 2024-02-24
Fix newline issue

pac48:
Newlines were breaking the code generation for Python https://github.com/PickNikRobotics/generate_parameter_library/issues/170. This PR fixes the issue and reformats the C++ to print better with `ros2 param describe...` The following out shows the new format.
```
Parameter name: hover_override
  Type: integer
  Description: Override hover action:
    0: Hover
    1: Push
    2: Pull
    -1: Do not override
  Constraints:
```
The old format was like this 
```
Parameter name: hover_override
  Type: integer
  Description: Override hover action: 0: Hover 1: Push 2: Pull -1: Do not override
  Constraints:
```

### DIFF
--- a/generate_parameter_library_py/generate_parameter_library_py/string_filters_cpp.py
+++ b/generate_parameter_library_py/generate_parameter_library_py/string_filters_cpp.py
@@ -10,7 +10,10 @@ def valid_string_cpp(description):
       str: The filtered string that is a valid C++ string.
     """
     if description:
-        filtered_description = description.replace('\n', '\\n    ')
+        # remove possible markdown/rst syntax, but add proper indent for cpp-header files.
+        filtered_description = (
+            description.replace('\\', '\\\\').replace('`', '').replace('\n', '\\n    ')
+        )
         return f'"{filtered_description}"'
     else:
         return '""'
--- a/generate_parameter_library_py/generate_parameter_library_py/string_filters_cpp.py
+++ b/generate_parameter_library_py/generate_parameter_library_py/string_filters_cpp.py
@@ -10,14 +10,8 @@ def valid_string_cpp(description):
       str: The filtered string that is a valid C++ string.
     """
     if description:
-        filtered_description = (
-            description.replace('\\', '\\\\').replace('"', '\\"').replace('`', '')
-        )
-        # create a quote delimited string for every line
-        filtered_description = '\n'.join(
-            f'"{line}"' for line in filtered_description.splitlines()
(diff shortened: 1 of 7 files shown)


christophfroehlich (approved):
Looks fine for me. 

Fyi, I have to change the parameter description of joint_trajectory_controller because I used `\pi` inside the description for rendering with sphinx -> without replacement of all `\` it throws warnings now. But it still works, and output of `ros2 param describe` looks way better.

pac48:
> Looks fine for me.
> 
> Fyi, I have to change the parameter description of joint_trajectory_controller because I used `\pi` inside the description for rendering with sphinx -> without replacement of all `\` it throws warnings now. But it still works, and output of `ros2 param describe` looks way better.

@christophfroehlich Do you mind giving this PR an approval review once you have made the changes you need to make to the joint_trajectory_controller?

generate_parameter_library_py/generate_parameter_library_py/string_filters_cpp.py
```diff
@@ -10,13 +10,23 @@ def valid_string_cpp(description):
...
     if description:
-        filtered_description = (
-            description.replace('\\', '\\\\').replace('"', '\\"').replace('`', '')
-        )
-        # create a quote delimited string for every line
-        filtered_description = '\n'.join(
-            f'"{line}"' for line in filtered_description.splitlines()
-        )
-        return filtered_description
+        filtered_description = description.replace('\n', '\\n    ')
```
christophfroehlich, line 13: Or this should be fine too? This would fix your indent and also the markdown syntax I added to [JTC](https://github.com/ros-controls/ros2_controllers/blob/master/joint_trajectory_controller/src/joint_trajectory_controller_parameters.yaml). 
```suggestion
        # remove possible markdown/rst syntax, but add proper indent for cpp-header files. 
        filtered_description = description.replace('\\', '\\\\').replace('"', '\\"').replace('`', '').replace('\\n', '\\n    ')
```

pac48: I added this change but made on small change. It should replace `\n' with `\\n' for the new liens to render correctly.

christophfroehlich (approved):
I tested it with joint_trajectory_controller, it works perfectly now.

sea-bass (approved):

saikishor:
Hello!

Thank you for fixing it. When can we expect to have this fix released upstream?

PR #183, opened 2024-03-19
Fix mapped paramteres generation for python

Kotochleb:
This PR fixes incorrectly created parameters for maps in Python.

With this yaml:
```yaml
my_node:
  my_params:
    params_to_be_mapped: {
      type: string_array,
    }
    __map_params_to_mapped:
      in_map_param: {
        type: bool,
      }
```

Current Jinja template will create following code:

```python
# declare any new dynamic parameters
for value_1 in updated_params.params_to_mapped:
  updated_params.my_params.add_entry(value_1)

# ...

for value_1 in updated_params.params_to_mapped:
  param_name = f"{self.prefix_}my_params.{value_1}.in_map_param"

# ...

# declare and set all dynamic parameters
for value_1 in updated_params.params_to_mapped:
  updated_params.my_params.add_entry(value_1)
```

While the expected output is:
```python
# declare any new dynamic parameters
for value_1 in updated_params.my_params.params_to_mapped:
  updated_params.my_params.add_entry(value_1)

# ...
for value_1 in updated_params.my_params.params_to_mapped:
  param_name = f"{self.prefix_}my_params.{value_1}.in_map_param"

# ...

# declare and set all dynamic parameters
for value_1 in updated_params.my_params.params_to_mapped:
  updated_params.my_params.add_entry(value_1)
```

Namespace of `my_params` is not taken into account in `for` loops of the mapped values.


sea-bass (commented):
Similarly to #185, this also will need to handle the case where `{{struct_name}}` does not actually exist.

Kotochleb:
@sea-bass I saw #185, but it did not solve this issue. I tested it once again after you comment and neither [edfeb82](https://github.com/PickNikRobotics/generate_parameter_library/commit/edfeb82ffcbb958785ee4f2d7b6eea38ee79dc78) nor current main do not work. That is why I still want to keep this PR open.

As you suggested, I added check if `{{struct_name}}` is not empty, and I also merged main to make sure everything will still work after adding #185.

example_python/generate_parameter_module_example/admittance_parameters.py
```diff
@@ -0,0 +1,1275 @@
+# flake8: noqa
```
pac48, line 1: Can you delete this autogenerated file from this PR?

Kotochleb: Removed. I don't know how it slipped in there. Thanks for the catch

pac48:
@Kotochleb Can you share the `parameters.yaml` file that caused your issue in the first place?

Kotochleb:
Thank you for clarification. Since the changes I proposed are breaking, I will simply modify my code to match current expectations. Feel free to close this PR without merging.

pac48:
@Kotochleb Do you mind opening an issue for this? That way, this feature can potentially be added when I have some time.

Kotochleb:
Since the issue is open now I am closing this PR

Issue #182, opened 2024-03-18
Issues building mapped data


sea-bass:
As described in https://github.com/ros-planning/moveit2/issues/2749, I think there is a bug introduced by #166.

Seems like this YAML description generates invalid C++ code:

```yaml
  __map_joints:
    weight: {
      type: double,
      default_value: 1.0,
      description: "Joint weight",
      validation: {
        gt<>: [ 0.0 ]
      }
    }
```

As the build errors for this are as follows:

```
--- stderr: moveit_kinematics
  In file included from /home/runner/work/moveit2/moveit2/.work/target_ws/src/moveit2/moveit_kinematics/kdl_kinematics_plugin/include/moveit/kdl_kinematics_plugin/kdl_kinematics_plugin.h:42,
                   from /home/runner/work/moveit2/moveit2/.work/target_ws/src/moveit2/moveit_kinematics/kdl_kinematics_plugin/src/kdl_kinematics_plugin.cpp:37:
  /home/runner/work/moveit2/moveit2/.work/target_ws/build/moveit_kinematics/kdl_kinematics_plugin/kdl_kinematics_parameters/include/kdl_kinematics_parameters.hpp: In member function ‘void kdl_kinematics::ParamListener::refresh_dynamic_parameters()’:
  /home/runner/work/moveit2/moveit2/.work/target_ws/build/moveit_kinematics/kdl_kinematics_plugin/kdl_kinematics_parameters/include/kdl_kinematics_parameters.hpp:146:36: error: expected unqualified-id before ‘.’ token
    146 |       auto& entry = updated_params..joints_map[value_1];
        |                                    ^
  /home/runner/work/moveit2/moveit2/.work/target_ws/build/moveit_kinematics/kdl_kinematics_plugin/kdl_kinematics_parameters/include/kdl_kinematics_parameters.hpp: In member function ‘rcl_interfaces::msg::SetParametersResult kdl_kinematics::ParamListener::update(const std::vector<rclcpp::Parameter>&)’:
  /home/runner/work/moveit2/moveit2/.work/target_ws/build/moveit_kinematics/kdl_kinematics_plugin/kdl_kinematics_parameters/include/kdl_kinematics_parameters.hpp:217:32: error: expected unqualified-id before ‘.’ token
    217 |                 updated_params..joints_map[value_1].weight = param.as_double();
        |                                ^
[...]

PR #185, opened 2024-03-20
Restore functionality for mapped params with no struct name

sea-bass:
This is... not pretty, but it seems to restore the functionality reported in https://github.com/PickNikRobotics/generate_parameter_library/issues/182 without breaking anything else.

@pac48 what do you think? More trouble than it's worth / should this just not be supported?

```yaml
  __map_joints:
    weight: {
      type: double,
      default_value: 1.0,
      description: "map parameter without struct name",
      validation: {
        gt<>: [0.0],
      }
    }
```

it also supports nested!

```yaml
  __map_joints:
    __map_dof_names:
      weight: {
        type: double,
        default_value: 1.0,
        description: "map parameter without struct name",
        validation: {
          gt<>: [0.0],
        }
      }
```


pac48 (approved):
Thanks for making this fix. I noticed CI is failing in rolling, but that seems to be unrelated.

PR #187, opened 2024-03-22
Fix test on rolling CI

pac48:
It appears that the CI for rolling might be running Python with optimization. This disables type checks causing a test to fail. This PR disables the test if the environment is set to optimize.


christophfroehlich (approved):
seems to be fine, but I don't have that much experience with python + CI

PR #194, opened 2024-04-22
Update black lib into .pre-commit-config.yaml

mosfet80:
Updated psf black to last 23 version .
Lot ox fix.
Python 3.7 runtime support are removed.
[docker] Build with 3.12 image


christophfroehlich:
Fyi: you could bump all hooks to the latest version by running `pre-commit autoupdate`.
I also wrote a [reusable workflow](https://github.com/ros-controls/ros2_control_ci/blob/master/.github/workflows/reusable-update-pre-commit.yml) for automating that, e.g., running it once a month.

christophfroehlich (approved):

pac48 (approved):

PR #203, opened 2024-05-03
Drop yaml brackets for consistency and readability

Timple:
I noticed in some reviews internally that the brackets were used inconsistently. Turns out they were mostly copy pasted from the great parameter library.

May I suggest dropping the bracket syntax? It's yaml after-all, not json :slightly_smiling_face: 

As you can already see in the first README example, it was a bit inconsistent, they are not used after `turtlesim` but are used after `background`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,30 +22,24 @@ Write a yaml file to declare your parameters and their attributes.
 ```yaml
 turtlesim:
   background:
-    r: {
-      type: int,
-      default_value: 0,
-      description: "Red color value for the background, 8-bit",
-      validation: {
+    r:
+      type: int
+      default_value: 0
+      description: "Red color value for the background, 8-bit"
+      validation:
         bounds<>: [0, 255]
-      }
-    }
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -310,7 +300,7 @@ tl::expected<void, std::string> integer_equal_value(
 ```
 To configure a parameter to be validated with the custom validator function `integer_equal_value` with an `expected_value` of `3` you could would this to the YAML.
 ```yaml
-validation: {
+validation:
   "my_project::integer_equal_value": [3]
 }
```
christophfroehlich, line 305:
```suggestion
```

christophfroehlich (approved):

README.md
Timple: Ah, well spotted!

Timple:
So after the approval, what's next? :slightly_smiling_face:

pac48 (approved):

PR #205, opened 2024-06-13
Add Non-Blocking try_get_params Function for Real-Time Control Systems

KentaKato:
This pull request adds a new function, `try_get_params()`, to improve responsiveness especially in real-time systems by avoiding blocking operations. The existing `is_old()` function also risks blocking, so I incorporated it into `try_get_params()`.

### DIFF
--- a/generate_parameter_library_py/generate_parameter_library_py/jinja_templates/cpp/parameter_library_header
+++ b/generate_parameter_library_py/generate_parameter_library_py/jinja_templates/cpp/parameter_library_header
@@ -123,6 +123,17 @@ struct StackParams {
       return params_;
     }
 
+    bool try_get_params(Params & params_in) const{
+      if (mutex_.try_lock()) {
+        if (const bool is_old = params_in.__stamp != params_.__stamp; is_old) {
+          params_in = params_;
+        }
+        mutex_.unlock();
+        return true;
+      }
+      return false;
+    }
+
     bool is_old(Params const& other) const {
       std::lock_guard<std::mutex> lock(mutex_);
       return params_.__stamp != other.__stamp;
--- a/generate_parameter_library_py/generate_parameter_library_py/jinja_templates/cpp/parameter_library_header
+++ b/generate_parameter_library_py/generate_parameter_library_py/jinja_templates/cpp/parameter_library_header
@@ -123,7 +123,7 @@ struct StackParams {
       return params_;
     }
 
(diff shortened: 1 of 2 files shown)


generate_parameter_library_py/generate_parameter_library_py/jinja_templates/cpp/parameter_library_header
```diff
@@ -123,6 +123,17 @@ struct StackParams {
       return params_;
     }
 
+    bool try_get_params(Params & params_in) const{
```
pac48, line 126:
```suggestion
    bool try_get_params(Params & params_in) const {
```

KentaKato: I was unaware of the "Suggested changes" feature, so I responded by committing directly to your suggestions.

pac48:
@KentaKato Is there any reason to return a bool and take a reference argument as opposed to returning an optional? I think memory will be allocated in either case.

KentaKato:
@pac48 
Thank you for your review.

When returning an `std::optional`, users can interact with the results in the following ways:

**Pattern 1:** This incurs a copy cost.
```cpp
if (const auto params_opt = try_get_params(); params_opt.has_value()) {
    params_local = params_opt.value();
}
```

**Pattern 2:** This avoids copy costs but is slightly more cumbersome.
```cpp
if (const auto params_opt = try_get_params(); params_opt.has_value()) {
    params_local = std::move(params_opt.value());
}
```

Given that the approach to return a bool and take a reference argument excels in both cost efficiency and ease of use, I opted for this method.

However, if returning an optional is preferred, I am open to making that change.
If we decide to return an optional, would the following logic be suitable? This allows users to detect when try_lock() fails:

1. try_lock() == true && is_old == true
    - return params_
2. try_lock() == true && is_old == false
    - return params_
3. try_lock() == false
    - return std::nullopt

(Cases 1 and 2 can be combined, resulting in returning params_ whenever try_lock() == true.)

pac48:
@KentaKato I think the pattern is okay. Since in the context of real time control, it is probably better. Can you ad a simple test that calls this function, maybe here example/test/example_test_gtest.cpp?

KentaKato:
@pac48 I've added the requested test in example/test/example_test_gtest.cpp. Please let me know if there's anything else that needs to be addressed. Thank you!

pac48 (approved):

PR #211, opened 2024-07-17
Add ability for param_listener to load values from yaml dictionary

### LINKED ISSUE #208: Set parameters from YAML dictionary.
I'm in the need of using generate_parameter_library to declare, set and update my parameters when launching/running my node but I also need the ability to control which values are loaded when testing.

When I test I create a ROS2 node, declare the ParamListener and then get my parameters but the problem is that nothing has set them so everything defaults. I would like the ability to get my parameters and then override them from a YAML dictionary

```python
self.node = rclpy.create_node('my_test_node')
self.param_listener = my_params.ParamListener(self.node)
self.params = self.param_listener.get_params()

path_to_test_config = os.path.dirname(__file__) + '/config/test_parameters_from_file.yaml'                   
# Load the parameters specific to the test Node
with open(path_to_test_config, 'r') as file:
     my_manually_loaded_yaml_parameters = yaml.safe_load(file)['my_test_node']['ros__parameters'] 
     print(my_manually_loaded_yaml_parameters['load_pose_list_from_file']) # print a parameter from the dictionary
```

If I could update the parameters from a YAML file it would allow me to write test configs that I can use to check expected behavior.
```python
self.node = rclpy.create_node('my_test_node')
self.param_listener = my_params.ParamListener(self.node)
path_to_test_config = os.path.dirname(__file__) + '/config/test_parameters_from_file.yaml'                   
# Update parameters from test config YAML file before fetching a copy.
with open(path_to_test_config, 'r') as file:
     my_manually_loaded_yaml_parameters = yaml.safe_load(file)['my_test_node']['ros__parameters'] 
     self.param_listener.update_internal_params(my_manually_loaded_yaml_parameters)
self.params = self.param_listener.get_params()
```

@pac48 does this sound hard to implement?

MarqRazz:
This PR makes it so the `parameter_listener` can refresh it's copy of parameters with ones specified in a yaml dictionary and would close #208.

To load and the get the parameters
```python
        self.param_listener = my_param_namespace.ParamListener(self.node)
        path_to_config = os.path.dirname(__file__) + '/config/my_config_file.yaml'
        with open(path_to_config) as file:
            configParamsFromYaml = yaml.safe_load(file)['my_param_namespace']['ros__parameters']

        self.param_listener.set_params_from_dict(configParamsFromYaml)
        self.test_params = self.param_listener.get_params()
```

### DIFF
--- a/generate_parameter_library_py/generate_parameter_library_py/jinja_templates/python/parameter_library_header
+++ b/generate_parameter_library_py/generate_parameter_library_py/jinja_templates/python/parameter_library_header
@@ -29,7 +29,7 @@ def unpack_parameter_dict(namespace: str, parameter_dict: dict):
     for param_name, param_value in parameter_dict.items():
         full_param_name = namespace + param_name
         # Unroll nested parameters
-        if type(param_value) == dict:
+        if isinstance(param_value, dict):
             nested_params = unpack_parameter_dict(
                     namespace=full_param_name + '.',
                     parameter_dict=param_value)
@@ -51,6 +51,7 @@ stamp_ = Time()
 {{sub_struct_content-}}
 {% endfilter -%}
 {%- endfilter %}
+
 
     class ParamListener:
         def __init__(self, node, prefix=""):
--- a/generate_parameter_library_py/generate_parameter_library_py/jinja_templates/python/parameter_library_header
+++ b/generate_parameter_library_py/generate_parameter_library_py/jinja_templates/python/parameter_library_header
@@ -15,29 +15,7 @@ from generate_parameter_library_py.python_validators import ParameterValidators
 
 {% if user_validation_file|length -%}
 import {{user_validation_file}} as custom_validators
-{% endif -%}
-
-def unpack_parameter_dict(namespace: str, parameter_dict: dict):
-    """
-    Flatten a parameter dictionary recursively.
-
-    :param namespace: The namespace to prepend to the parameter names.
-    :param parameter_dict: A dictionary of parameters keyed by the parameter names
-    :return: A list of rclpy Parameter objects
-    """
-    parameters = []
-    for param_name, param_value in parameter_dict.items():
-        full_param_name = namespace + param_name
-        # Unroll nested parameters
-        if isinstance(param_value, dict):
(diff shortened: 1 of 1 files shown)


MarqRazz:
Note this design was heavily influenced by rclpy.parameter.py's [unpack_parameter_dict()](https://github.com/ros2/rclpy/blob/jazzy/rclpy/rclpy/parameter.py#L328-L349)

What do you think about supporting the yaml loading like they do in [parameter_dict_from_yaml_file()](https://github.com/ros2/rclpy/blob/jazzy/rclpy/rclpy/parameter.py#L276). I tried to use this functionality but it's not in Humble and returns a `Dict[str, ParameterMsg]` where generate_parameter's `update()` takes a list of `rclpy.parameter.Parameter`'s. Personally I think it's fine for the user to have to load the dictionary from the correct namespace with `yaml.safe_load(file)['my_param_namespace']['ros__parameters']`

MarqRazz:
@pac48 do you have any thoughts about this approach? Is there anything I can do to help get this across the finish line?

generate_parameter_library_py/generate_parameter_library_py/jinja_templates/python/parameter_library_header
```diff
@@ -10,11 +10,34 @@ from rclpy.exceptions import InvalidParameterValueException
...
+        full_param_name = namespace + param_name
+        # Unroll nested parameters
+        if isinstance(param_value, dict):
+            nested_params = unpack_parameter_dict(
+                    namespace=full_param_name + rclpy.parameter.PARAMETER_SEPARATOR_STRING,
+                    parameter_dict=param_value)
+            parameters.extend(nested_params)
+        else:
+            parameters.append(rclpy.parameter.Parameter(full_param_name, value=param_value))
+    return parameters
```
pac48, line 39: Can we move this function into the class as a method?

pac48:
@MarqRazz I added a small comment, otherwise looks good.

MarqRazz:
How does that look?

MarqRazz:
@pac48 can we run the tests?

pac48 (approved):

Issue #204, opened 2024-05-30
How to include the parameters in a dependent package?


reinzor:
Hi,

Thanks for this library!

We would like to include a public header from another packages that includes the parameter header. However, this header is not available for that package as `#include <example_parameters.hpp>` but it is installed under `example_parameters/example_parameters.hpp`. How can we do this?

So the setup:

package_with_parameters/include/header.hpp

```
#include <example_parameters.hpp>
```

another_package/include/header.hpp

```
#include <package_with_parameters/header.hpp> // #include <example_parameters.hpp> cannot be found, example_parameters/example_parameters.hpp can be found
```

Can you help us with this?

christophfroehlich:
I found a way to install the created headers
```CMake
install(FILES ${CMAKE_CURRENT_BINARY_DIR}/limiters_parameters/include/limiters_parameters.hpp
  DESTINATION include/${PROJECT_NAME}/${PROJECT_NAME})
```
and now another package can 

```c++
#include "steering_controllers_library/limiters_parameters.hpp"
```

But there is for sure a smarter way. Maybe someone with knowledge in CMake can enlighten us ;)

reinzor:
Thanks for your quick response.

Including the parameter file via another package already works. The problem here is that the include is different for the package that holds defines the parameters and a dependent package. This include:

```cpp
#include "steering_controllers_library/limiters_parameters.hpp"
```

is not being resolved correctly by the package itself. There, it is available as

```cpp
#include "limiters_parameters.hpp"
```

Or am I missing something?

christophfroehlich:
That's how I use it now.
but :eyes: https://github.com/PickNikRobotics/generate_parameter_library/issues/180

reinzor:
Thanks! This would indeed allow for including a header that includes the parameter header in another package.

AugusteBourgois:
Hi everyone, PR #213 should fix the issue. It's on a fork right now, but can hopefully be merged soon. Open to discussion anyway.

Issue #180, opened 2024-03-06
Changing Parameter Library Header Installation Location


jcarpinelli-bdai:
Is it possible to change the location of installed parameter library header files? For example, in your [example](https://github.com/PickNikRobotics/generate_parameter_library/blob/4b50b985447ac2b20e6b54eb695ea11725a19d89/example/include/generate_parameter_library_example/minimal_publisher.hpp#L34) you include the parameter file header without a folder prefix. Is there any way to change this library's header install directory? I am trying to remove the `<parameter_library_name>/include` prefix, so I can `#include "parent_package/parameter_library_header.hpp`.

Timple:
This would help us too! :slightly_smiling_face:

AugusteBourgois:
Hi everyone, PR https://github.com/PickNikRobotics/generate_parameter_library/pull/213 should fix the issue. It's on a fork right now, but can hopefully be merged soon. Open to discussion anyway.

PR #213, opened 2024-08-02
Change header install path

AugusteBourgois:
Hi everyone,

Thanks for the great work behind generate_parameter_library ! It definitely spares us many lines of code.

A problem met by myself and some users according to #180 and #204 is that the generated header cannot be easily used outside the package that generated them. So I've made a small change to fix this. I'm open to discuss this more in depth if necessary, and add tests if required.

This PR solves #180 and #204 by installed the generated header in the same include folder as the current package, so that it can be used in other packages, and not only in the current one.


christophfroehlich (changes requested):
Thanks for working on this. I tried now to use your patch. The idea is then to include this with 
```cpp
#include "gpl_cmake_target_name.hpp"
```
in the same package as well as in dependent packages?

Maybe
```cpp
#include "package_name/gpl_cmake_target_name.hpp"
```
would be a better choice? (or both, otherwise existing packages will break).

Currently, the dependencies of GPL are not exported (rsl + parameter_traits). A simple find_package of the package including the package creating the GPL-library is not sufficient. Could they be exported as well?

AugusteBourgois:
Hi, thanks for the feedback. I pushed a little patch as I noticed this first version did not work in all cases.

Now, say you have `package1` generating a header `param_file.hpp`.

You need to link you libs and executables against the generated target to access the include file, and export it so that it can be accessed from other packages. Finally, you need to add `generate_package_library` to your `ament_export_dependencies` so that other packages have access to GPL's dependencies (rsl, package_traits...). I guess a `find_package` and ament_export_dependencies` could work, but I'm not sure.

```cmake

find_package(ament_cmake REQUIRED)
find_package(generate_package_library REQUIRED)

generate_package_library(param_lib path_to_param_file.hpp)

# compiled library
add_library(lib1 PUBLIC lib1.cpp)
target_include_directories(lib1 PUBLIC
    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
    $<INSTALL_INTERFACE:include>)
target_link_libraries(lib1 PUBLIC param_lib)

# header only library
add_library(lib2 INTERFACE)
target_include_directories(lib2 INTERFACE
    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
    $<INSTALL_INTERFACE:include>)
target_link_libraries(lib2 PUBLIC param_lib)

# export targets
install(TARGETS lib1 lib2 param_lib EXPORT ${PROJECT_NAME}Targets)

ament_export_targets(${PROJECT_NAME}Targets HAS_LIBRARY_TARGET)
ament_export_dependencies(generate_parameter_library ...)
ament_package()
```

The generated header can be included in `lib1.cpp` and `lib2.hpp` using 
```cpp
#include <package1/param_file.hpp>
```
 The include line will be the same for other packages.

The current solution breaks existing packages. I can improve it so that it does not, at the cost of duplicating the generated header in the build folder (seems fine to me). I can also try to avoid exporting generate_parameter_library as a dependency, which, I agree, would be ideal.

Any thoughts ?

christophfroehlich:
> You need to link you libs and executables against the generated target to access the include file, and export it so that it can be accessed from other packages. Finally, you need to add `generate_package_library` to your `ament_export_dependencies` so that other packages have access to GPL's dependencies (rsl, package_traits...). I guess a `find_package` and ament_export_dependencies` could work, but I'm not sure.
> 
> ```cmake
> find_package(ament_cmake REQUIRED)
> find_package(generate_package_library REQUIRED)
> 
> generate_package_library(param_lib path_to_param_file.hpp)
> 
> # compiled library
> add_library(lib1 PUBLIC lib1.cpp)
> target_include_directories(lib1 PUBLIC
>     $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
>     $<INSTALL_INTERFACE:include>)
> target_link_libraries(lib1 PUBLIC param_lib)
> 
> # header only library
> add_library(lib2 INTERFACE)
> target_include_directories(lib2 INTERFACE
>     $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
>     $<INSTALL_INTERFACE:include>)
> target_link_libraries(lib2 PUBLIC param_lib)
> 
> # export targets
> install(TARGETS lib1 lib2 param_lib EXPORT ${PROJECT_NAME}Targets)
> 
> ament_export_targets(${PROJECT_NAME}Targets HAS_LIBRARY_TARGET)
> ament_export_dependencies(generate_parameter_library ...)
> ament_package()
> ```

The easier it is to use generate_parameter_library, the better. If there is a way to avoid additional cmake stuff in the library, which generates the code, I would vote for it. But maybe there is an argument to not exporting it, if it is not reused in a different package -> then I'm fine for a manual export step together with clear instructions in the docs (which is missing now anyways).

> The generated header can be included in `lib1.cpp` and `lib2.hpp` using
> 
> ```c++
> #include <package1/param_file.hpp>
> ```
> 
> The include line will be the same for other packages.

This is great.

> The current solution breaks existing packages. I can improve it so that it does not, at the cost of duplicating the generated header in the build folder (seems fine to me). I can also try to avoid exporting generate_parameter_library as a dependency, which, I agree, would be ideal.

Could we deprecate the files in the "old" path with some compile-time warning? and then remove that after a couple of releases?

AugusteBourgois:
My last push :

- installs the generated header (and validation header) so that they are accessible via `#include <package_name/parameter_file.hpp>` from inside and outside the package.
- copies the generated header (and validation header) so they they remain accessible via `#include "parameter_file.hpp"` as they used to from inside the same package.
- deprecates the local generated header, so that a warning is printed at build time.
- exports the generated header dependencies (fmt...) so that there is no need for `ament_export_dependencies(generate_parameter_library)` anymore.

In the next push, I will propose an updated documentation and an extra example package using the generated header from generate_parameter_library_example.

The only thing I haven't tried is the python version, since we don't use python at the moment.

README.md
```diff
@@ -348,10 +348,46 @@ The generated parameter value for the nested map example can then be accessed wi
...
+
+install(DIRECTORY include/${PROJECT_NAME}/ DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/${PROJECT_NAME})
+
+install(TARGETS my_lib
+  EXPORT ${PROJECT_NAME}Targets
+  ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}
+  LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
+  RUNTIME DESTINATION lib/${PROJECT_NAME})
+
+ament_export_targets(${PROJECT_NAME}Targets HAS_LIBRARY_TARGET)
```
christophfroehlich, line 386: The equivalent of these two instructions have to be added in the library, which exports the turtlesim_parameters. Please add this to this section here (it is not part of the [minimum CMakeLists snippet](https://github.com/PickNikRobotics/generate_parameter_library/blob/main/README.md#add-parameter-library-generation-to-project))

christophfroehlich (changes requested):
Thanks for the followup, I just tested this within my projects:
It works as described, but I had to fight CMake to export that as expected (see my comments below).

Furthermore, the include paths in the old examples have to be updated:
```
from src/generate_parameter_library/example/src/minimal_publisher.cpp:29:
/build/generate_parameter_library_example/include/admittance_controller_parameters.hpp:1:179: note: ‘#pragma message: #include "admittance_controller_parameters.hpp" is deprecated. Use #include <generate_parameter_library_example/admittance_controller_parameters.hpp> instead.’
    1 | #pragma message("#include \"admittance_controller_parameters.hpp\" is deprecated. Use #include <generate_parameter_library_example/admittance_controller_parameters.hpp> instead.")
```

example_external/CHANGELOG.rst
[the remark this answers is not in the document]
AugusteBourgois: Ok, removed the content of the file. Also corrected the old headers.

README.md
christophfroehlich: would there be any drawbacks?

AugusteBourgois: The only effect would be to export the generated header target even if you don't use it outside of your package, which sounds fine to me (same thing with exported dependencies). The main advantage is to group everything under the generate_parameter_library cmake command, which makes it very easy to use.

christophfroehlich: again, I vote for the simplest version to use ;) but I'd understand if someone wants to keep the created artifacts as small as possible

AugusteBourgois: Done. To achieve this, I transformed the function into a macro, so that I don't need to globalize every cmake variable set in the function.

AugusteBourgois:
Hi @christophfroehlich, I have pushed something to correct the cmake linter error. colcon test now runs without errors, at least locally.

README.md
```diff
@@ -63,6 +63,9 @@ target_link_libraries(minimal_node PRIVATE
   rclcpp::rclcpp
   turtlesim_parameters
 )
+
+install(TARGETS minimal_node turtlesim_parameters
+  EXPORT ${PROJECT_NAME}Targets
```
christophfroehlich, line 68: there is a closing parenthesis missing.

Could you explain the difference of 
`EXPORT ${PROJECT_NAME}Targets` here
and 
`EXPORT export_generate_parameter_library_example` from the example/CMakeLists.txt please?

christophfroehlich (commented):
I just have tested this again. Could you comment on the changes of your last commit? Where is the problem with a breaking change?

I needed to add the equivalent of 

```
install(TARGETS ${LIB_NAME} EXPORT ${PROJECT_NAME}Targets)
ament_export_targets(${PROJECT_NAME}Targets HAS_LIBRARY_TARGET)
```
to my code to get that exported, the snippet from the README without ament_export_targets() did not work for me.

README.md
AugusteBourgois: Hi,

So the reason why I made these changes is that we tried to package many ros2 packages, and that some failed because of the changes I made in the first place.

The problem with my first implementation was that the generated parameters target was exported automatically by the cmake macro under the name `${PROJECT_NAME}Targets`. Say you had another target, like a library, using the generated parameters one, and exported under a different name than `${PROJECT_NAME}Targets`, cmake would raise an error saying that the generated parameter target was exported multiple times.

So the solution I found was to remove the automatic target export from the macro. So instead of
```cmake
generate_parameters(params_target params_file.yaml)

add_library(my_lib my_lib.cpp)
target_link_libraries(my_lib SHARED params_target)

install(TARGETS my_lib EXPORT WhatheverExportNameYouWant)
ament_export_targets(WhatheverExportNameYouWant HAS_LIBRARY_TARGET)
```

You now need to specify `params_target` in the `TARGETS` list.
This actually make sense : similarly to `add_library`, the `generate_parameters` macro creates a cmake target. It is up to you to export it if you want, and under whatever name you want.

To answer your last question, these are simply names. You can write whatever you want. You could export multiple targets under different names I guess, although in small packages you tend to export all targets under a single name. At least that's what I encountered so far. `${PROJECT_NAME}Targets` will simply expand to `MyProjectTargets` if your cmake project is called `MyProject`.

To conclude, this new implementation requires slightly more work since you need to export the generated parameters target manually if you want to use it outside of your package. But it does not break current implementation in many packages such as `ros2_control`, which is better if you want this path to be released quickly, and backported (no additional work required) to humble.

Do not hesitate if you have more questions or comments.

AugusteBourgois: Hi @christophfroehlich 
Is there something else you would like to address regarding this PR ?

christophfroehlich: Sorry, I haven't found time to test the last version. But there is a merge conflict now with the main branch?

AugusteBourgois: I rebased the branch on the last version of main. All tests now pass.

christophfroehlich (approved):
Thanks for your efforts, I have tested that now successfully and would love to use that in ros2_control soon :+1: 

Only one question: Is it necessary to link the parameters from the external library? I did not see any difference if I don't link it against the lib.
`target_link_libraries(my_lib PUBLIC turtlesim::turtlesim_parameters)`

AugusteBourgois:
Great ! Thank you for your review :)

I'm not sure I understand your use case. Can you please give me more information about what you are trying to do ?

example_external/CMakeLists.txt
```diff
@@ -0,0 +1,52 @@
...
+)
+target_include_directories(minimal_publisher_external PUBLIC
+  $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>
+  $<INSTALL_INTERFACE:include/generate_parameter_library_example_external>
+)
+target_link_libraries(minimal_publisher_external
+  PUBLIC
+    rclcpp::rclcpp
+    rclcpp_components::component
+    generate_parameter_library_example::admittance_controller_parameters
```
christophfroehlich, line 26: I'm just referring to the external example you provided: I did the same but not added it to the target_link_libraries, and it builds fine. (not sure about the implications with CMake and header-only libraries).

AugusteBourgois: If you comment out this line and try to build this package in a clean workspace, it should crash and say that it cannot find the parameters headers.

I haven't seen your example and the way dependencies are created and linked together, but my guess is that the generated parameter library is linked one way or another to your own library, preventing the crash.
Another guess is that since it is a header only library, so if you somehow indicate the headers' path in `target_include_directories`, the build won't crash.

AugusteBourgois: In other words, linking a target with an interface library comes down to adding the correct headers' path. This is the modern Cmake way of doing things, AFAIK.

sea-bass:
I can try do some testing soon.

Quick question for now: since this is a breaking change, is the intention to leave Humble (and maybe Jazzy) as-is, and land this to Rolling?

Or can/should this all just go into every active distro because of the copied file with deprecation warnings?

christophfroehlich:
At least it would be nice to have in jazzy.
I don't have a strong opinion about humble. The result looks nicer, and it is fully backwards compatible. So maybe its worth backporting it back to humble, as it is still an active distro for quite some time.

README.md
```diff
@@ -350,10 +354,46 @@ The generated parameter value for the nested map example can then be accessed wi
...
+#include <turtlesim/turtlesim_parameters.hpp>
 ```
+
+Note that this header can also be used from another package:
+```cmake
+cmake_minimum_required(VERSION 3.8)
+project(my_other_package)
+
+
+include(GNUInstallDirs)
```
sea-bass, line 368: is this standard in ROS boilerplate cmakelists.txtx?

example_external/README.md
```diff
@@ -0,0 +1,155 @@
+# Example:
```
sea-bass, line 1: This probably could use a more descriptive title showing this is for how to include parameters from another library, as well as a sentence or two describing things before jumping right into instructions.

example_external/README.md
```diff
@@ -0,0 +1,155 @@
...
+  qos_overrides./parameter_events.publisher.reliability
+  scientific_notation_num
+  state_interfaces
+  three_numbers
+  three_numbers_of_five
+  use_feedforward_commanded_input
+  use_sim_time
+  ```
+
+All parameter are automatically declared and callbacks are setup by default. You can set a parameter by typing:
```
sea-bass, line 96:
```suggestion
All parameters are automatically declared and callbacks are setup by default.
You can set a parameter by typing:
```

example_external/README.md
```diff
@@ -0,0 +1,155 @@
...
+
+`ros2 param set /admittance_controller control.frame.id new_frame`
+
+You should see:
+
+`[INFO] [1656019001.515820371] [admittance_controller]: New control frame parameter is: 'new_frame'`
+
+Congratulations, you updated the parameter!
+
+If you try to set a parameter that is read only, you will get an error. Running the following
```
sea-bass, line 106:
```suggestion
If you try to set a parameter that is read only, you will get an error.
Running the following
```

example_external/include/generate_parameter_library_example_external/minimal_publisher_external.hpp
```diff
@@ -0,0 +1,52 @@
+// Copyright 2022 PickNik Inc.
```
sea-bass, line 1: should be 2025 and you can feel free to attribute yourself instead!

example_external/src/minimal_publisher_external.cpp
```diff
@@ -0,0 +1,74 @@
+// Copyright 2022 PickNik Inc.
```
sea-bass, line 1: same here, 2025 and use whatever attributions makes sense for you

generate_parameter_library/cmake/generate_parameter_library.cmake
```diff
@@ -48,37 +48,46 @@ function(generate_parameter_library LIB_NAME YAML_FILE)
...
-      OUTPUT ${VALIDATE_HEADER}
-      COMMAND ${CMAKE_COMMAND} -E copy ${IN_VALIDATE_HEADER} ${VALIDATE_HEADER}
-      DEPENDS ${IN_VALIDATE_HEADER}
-      COMMENT
-      "Running `${CMAKE_COMMAND} -E copy ${IN_VALIDATE_HEADER} ${VALIDATE_HEADER}`"
-      VERBATIM
-    )
+    file(COPY ${IN_VALIDATE_HEADER} DESTINATION ${LIB_INCLUDE_DIR})
+    # necessary so that #include <param_file.hpp> can be used in the local package (deprecated)
+    file(COPY ${IN_VALIDATE_HEADER} DESTINATION ${CMAKE_CURRENT_BINARY_DIR}/include)
```
sea-bass, line 53: thank you for keeping this around for compatibility!

sea-bass (commented):
I have a few small comments and LGTM.

Based on looking in more detail, I'm on board with releasing with these changes to Humble and Jazzy as well.

@pac48 and/or @jcarpinelli-bdai -- do you guys want to chime in on this at all? I feel like I'm not close enough to this tool to be the final arbiter here.

README.md
AugusteBourgois: Not really, this is a standard cmake module allowing to use GNU standard installation directories instead of manually defined ones, through cmake variables. This is not necessary for a minimal example, but I would deem it a good practice.

AugusteBourgois:
Hi @sea-bass ,

Thank you for your thorough review. I will include your suggestions today.

Concerning the targeted release version, it is fully backwards compatible with Humble : it is being used on a daily basis in our Humble codebase.

generate_parameter_library/cmake/generate_parameter_library.cmake
AugusteBourgois: You're welcome !

README.md
sea-bass: Sounds good to me, thank you.

jcarpinelli-bdai:
> @pac48 and/or @jcarpinelli-bdai -- do you guys want to chime in on this at all? I feel like I'm not close enough to this tool to be the final arbiter here.

This all looks great to me. I'm familiar with this library as a user, and my only concern was backwards compatibility. Is the default header file location still accessible for libraries to `#include "*_parameters.hpp"`? From @AugusteBourgois's [comment](https://github.com/PickNikRobotics/generate_parameter_library/pull/213#issuecomment-2572641140) above, it seems that is the case. So if that's all true, I have no other comments. 

Thanks @AugusteBourgois for the contribution!

sea-bass (approved):
Alright, let's do this!

With this PR, I think it's fair that the next release should bump versions to `0.4.0`.

I'll do that and release to all the distros at the end of the week/weekend.

PR #214, opened 2024-08-07
Fix a typo preventing to use the full int64_t range when using gt<> and lt<> validators

### LINKED ISSUE #199: gt<> sets a maximum numeric limit to std::numeric_limits<int>::max() instead of std::numeric_limits<int64_t>::max()
Hi everyone,

I came across the following bug:

When using the gt<> parameter validator, I checked the source code generated, and it seems that when using gt<> on an integer (ros parameter type) the parameter created is of type int64_t, which is expected, but the maximum value is set to std::numeric_limits<int>::max().
There is a similar problem with lt<>.

I guess this is not the desired behavior.

This can be seen in the generated `admittance_controller_parameters.hpp` file, line 1072..1091 :

```cpp
if (!parameters_interface_->has_parameter(prefix_ + "lt_eq_fifteen")) {
          descriptor.description = "should be a number less than or equal to 15";
          descriptor.read_only = false;
          descriptor.integer_range.resize(1);
          descriptor.integer_range.at(0).from_value = std::numeric_limits<int>::lowest();
          descriptor.integer_range.at(0).to_value = 15;
          auto parameter = to_parameter_value(updated_params.lt_eq_fifteen);
          parameters_interface_->declare_parameter(prefix_ + "lt_eq_fifteen", parameter, descriptor);
      }
      if (!parameters_interface_->has_parameter(prefix_ + "gt_fifteen")) {
          descriptor.description = "should be a number greater than 15";
          descriptor.read_only = false;
          descriptor.integer_range.resize(1);
          descriptor.integer_range.at(0).from_value = 15;
          descriptor.integer_range.at(0).to_value = std::numeric_limits<int>::max();
          auto parameter = to_parameter_value(updated_params.gt_fifteen);
          parameters_interface_->declare_parameter(prefix_ + "gt_fifteen", parameter, descriptor);
      }
```

AugusteBourgois:
Hi everyone,

This PR fixes #199. Since ROS uses int64_t to store integer parameters, int64_t numeric limits should be used instead of int.

i simply corrected the template files and the associated unit tests.

Any thoughts ?

### DIFF
--- a/example/test/descriptor_test_gtest.cpp
+++ b/example/test/descriptor_test_gtest.cpp
@@ -79,14 +79,14 @@ TEST_F(DescriptorTest, check_lower_upper_bounds) {
 
 TEST_F(DescriptorTest, check_lt_eq) {
   EXPECT_EQ(descriptors_[3].integer_range.at(0).from_value,
-            std::numeric_limits<int>::lowest());
+            std::numeric_limits<int64_t>::lowest());
   EXPECT_EQ(descriptors_[3].integer_range.at(0).to_value, 15);
 }
 
 TEST_F(DescriptorTest, check_gt) {
   EXPECT_EQ(descriptors_[4].integer_range.at(0).from_value, 15);
   EXPECT_EQ(descriptors_[4].integer_range.at(0).to_value,
-            std::numeric_limits<int>::max());
+            std::numeric_limits<int64_t>::max());
 }
 
 int main(int argc, char** argv) {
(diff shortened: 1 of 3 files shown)


christophfroehlich (approved):
Sounds reasonable, thx for fixing.

AugusteBourgois:
My pleasure, happy to help.

pac48 (approved):

PR #218, opened 2024-09-09
Disable cache for `generate_parameter_library_cpp_BIN`

pac48:
The variable `generate_parameter_library_cpp_BIN` was cached, which caused build issues when switching between binary and source installed versions. This prevents the error `/opt/ros/humble/bin/generate_parameter_library_cpp: No such file or directory`.
`


sea-bass (approved):
Not sure if I still have any power for this approve to count, but...

Guess not rip

JafarAbdi (approved):

PR #220, opened 2024-09-19
fix error with not being able to find unpack_parameter_dict when using set_params_from_dict

### LINKED ISSUE #219: Usage of set_params_from_dict fails
When using the new feature to set params from a dict, the node fails with the following error:

`NameError: name 'unpack_parameter_dict' is not defined`

I think the error comes from here:

https://github.com/PickNikRobotics/generate_parameter_library/blob/28b187098a6ed12e6fff2e4276cd2b1f365b4e61/generate_parameter_library_py/generate_parameter_library_py/jinja_templates/python/parameter_library_header#L57-L80

The method is not referred to via self, and so python fails to find it.

Testing locally, simply removing @staticmethod and adding self seems to solve the issue.

Glonks:
This closes #219

### DIFF
--- a/generate_parameter_library_py/generate_parameter_library_py/jinja_templates/python/parameter_library_header
+++ b/generate_parameter_library_py/generate_parameter_library_py/jinja_templates/python/parameter_library_header
@@ -54,8 +54,7 @@ stamp_ = Time()
         def is_old(self, other_param):
             return self.params_.stamp_ != other_param.stamp_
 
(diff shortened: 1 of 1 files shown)


sea-bass (approved):
I guess another alternative to removing the `@staticmethod` is to change the calls to `ParamListener.unpack_parameter_dict(...)`. But this is good for me.

JafarAbdi (approved):

PR #221, opened 2024-10-10
Add "additional_constraints" support

MrBlenny:
This PR adds the ability to set "additional_constraints" in the [ParameterDescription](https://docs.ros2.org/galactic/api/rcl_interfaces/msg/ParameterDescriptor.html)


pac48:
@MrBlenny can you run `pre-commit run -a` in the root of your repository to fix the formatting?

The change looks nice overall! Is the idea of `additional_constraints` to pass the validation logic off to the downstream code? Since there might be domain specific validation, this seems useful.

MrBlenny:
I'll push a commit with lint when I'm at a computer later today.

Yes, I was keen to embed JSON schema into the additional constraints and use that to generate a UI for setting advanced parameters (such as parameters that have their values as JSON encoding ROS messages). A simpler example would be to embed enum options for a strong param in the additional constraints (as JSON schema or otherwise) so you can dynamically build a UI to set params via a dropdown of enum options.

It would be pretty powerful if this library supported JSON schema as a validator but that'd of course be a big change.

pac48 (approved):

PR #232, opened 2025-01-06
0.4.0

sea-bass:
Preparing the `0.4.0` release.

Will leave this for review for a little bit in case something new comes in, and will merge/bloom at the end of the week?

PRs to merge:
- [x] #236
- [x] #237 (or close it)
- [x] #238


example_python/setup.py
```diff
@@ -18,7 +18,7 @@
 
 setup(
     name=package_name,
-    version='0.3.8',
```
sea-bass, line 21: woops, I had done this wrong in 0.3.9 it seems 😅

generate_parameter_library_py/setup.py
```diff
@@ -6,7 +6,7 @@
 
 setup(
     name=package_name,
-    version='0.3.8',
```
sea-bass, line 9: same here lol

example_external/package.xml
```diff
@@ -2,7 +2,7 @@
 <?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
 <package format="3">
   <name>generate_parameter_library_example_external</name>
-  <version>0.1.0</version>
+  <version>0.4.0</version>
```
sea-bass, line 5: Note I changed this package to just track the same version as the rest of the repo.

I had overlooked that in reviewing the original PR.

example_external/CHANGELOG.rst
```diff
@@ -0,0 +1,68 @@
...
+------------------
+
+0.2.1 (2022-08-02)
+------------------
+
+0.2.0 (2022-08-01)
+------------------
+
+0.1.0 (2022-07-27)
+------------------
```
sea-bass, line 68: Would it work to remove all these other older releases from the file, since technically this package just popped up at 0.4.0?

christophfroehlich: good question, I don't have that much experience with the release process yet. 

for new packages in [ros2_controllers](https://github.com/ros-controls/ros2_controllers/blob/master/mecanum_drive_controller/CHANGELOG.rst) @bmagyar just left it as it is

christophfroehlich (approved):

example_external/CHANGELOG.rst
bmagyar: Can't help with that one. The way you get this is when running `catkin_generate_changelog --all` really goes through the history to match everything backwards. Not sure if there is anything on the buildfarm relying on this.

sea-bass: Thanks! I think I'm just going to start the history at 0.4.0.

sea-bass:
Do not merge yet!

I need to update these changelogs, which I will get to this evening.

PR #234, opened 2025-01-09
Update accessing mapped parameters in README

pawelir:
Currently an example of accessing mapped parameters is misleading. This PR aims to propose an update showcasing the exact syntax.

### DIFF
--- a/README.md
+++ b/README.md
@@ -350,7 +350,19 @@ cpp_name_space:
         type: double_array
 ```
 
-The generated parameter value for the nested map example can then be accessed with `params.gain.get_entry("joint1").get_entry("position").value`.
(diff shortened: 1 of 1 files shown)


christophfroehlich (changes requested):
but the `.at()` is the way to access them in C++, we use that in [ros2_controllers](https://github.com/ros-controls/ros2_controllers/blob/master/joint_trajectory_controller/src/joint_trajectory_controller.cpp#L1597)

`.get_entry()` is for python, so we should add both

pawelir:
Sure @christophfroehlich, please find updated version

christophfroehlich (approved):
Thanks!

sea-bass (approved):

Issue #233, opened 2025-01-07
ModuleNotFoundError python unit test


imcelroy:
Hello,

I am having errors when running `colcon test` for unit tests of python modules that use parameters generated by `generate_parameter_library`.

I have created a simple unit test that reproduces this error using your `example_python` on this branch: [bug/python-example-unit-test](https://github.com/forssea-robotics/generate_parameter_library/tree/bug/python-example-unit-test). It simply imports the `minimal_publisher` module.

Here are the steps to reproduce the error:

1. `rosdep update`
2. `rosdep install --rosdistro humble`
3. `colcon build --cmake-args -DBUILD_TESTING=1 -DCMAKE_EXPORT_COMPILE_COMMANDS=On --event-handlers console_direct+ --merge-install`
4. `colcon test --merge-install --packages-select generate_parameter_module_example --event-handlers console_cohesion+ log_command+ summary+ --return-code-on-test-failure`

The result should give something like the following:

```
==================================== ERRORS ====================================
________________________ ERROR collecting test session _________________________
/usr/lib/python3/dist-packages/pluggy/hooks.py:286: in __call__
    return self._hookexec(self, self.get_hookimpls(), kwargs)
/usr/lib/python3/dist-packages/pluggy/manager.py:92: in _hookexec
    return self._inner_hookexec(hook, methods, kwargs)
/usr/lib/python3/dist-packages/pluggy/manager.py:83: in <lambda>
    self._inner_hookexec = lambda hook, methods, kwargs: hook.multicall(
/usr/lib/python3/dist-packages/_pytest/python.py:200: in pytest_collect_file
    module: Module = ihook.pytest_pycollect_makemodule(path=path, parent=parent)
/usr/lib/python3/dist-packages/pluggy/hooks.py:286: in __call__
    return self._hookexec(self, self.get_hookimpls(), kwargs)
/usr/lib/python3/dist-packages/pluggy/manager.py:92: in _hookexec
    return self._inner_hookexec(hook, methods, kwargs)
/usr/lib/python3/dist-packages/pluggy/manager.py:83: in <lambda>
    self._inner_hookexec = lambda hook, methods, kwargs: hook.multicall(
/opt/ros/humble/lib/python3.10/site-packages/launch_testing/pytest/hooks.py:193: in pytest_pycollect_makemodule
    entrypoint = find_launch_test_entrypoint(path)
[...]

PR #236, opened 2025-01-11
Revert "Fix python lib install path (#207)"

sea-bass:
This reverts commit 4e92cc4e9da5e7a853ac0b1762db85e1f859d91b.

Turns out I took #207 in good faith, but it does cause problems.

Without that PR, this is what a custom module named `admittance_params.py` installs into. This is correct:

image

With that PR, it goes to the wrong folder one level up:

image

PR #237, opened 2025-01-11
Unit test friendly structure

sea-bass:
Attempts to address https://github.com/PickNikRobotics/generate_parameter_library/issues/233 by moving files around so that there is no module that could be wrongly imported in the source folder of the Python example package.


example_python/test/test_parameters.py
```diff
@@ -0,0 +1,7 @@
+import generate_parameter_module_example
+print(f'Imported file from: {generate_parameter_module_example.__file__}')
+print('OK 1')
+
+import generate_parameter_module_example.admittance_parameters  # noqa: E402
+
+print('OK 2')
```
sea-bass, line 7: note this test is only there for illustrative purposes -- we can either remove it or clean it up

generate_parameter_library_py/generate_parameter_library_py/setup_helper.py
```diff
@@ -55,6 +55,7 @@ def generate_parameter_module(module_name, yaml_file, validation_module=''):
             install_dir = os.path.join(
                 colcon_ws,
                 'install',
+                pkg_name,
```
sea-bass, line 58: to be fixed in #236

christophfroehlich (commented):
Sorry, I don't have much experience with python install stuff to review that :man_shrugging:

sea-bass:
> Sorry, I don't have much experience with python install stuff to review that 🤷‍♂️

No worries, I'm hoping the handful of people engaged in the associated issue will get a chance to test here. I just can't officially tag most of them as reviewers as they are not repo collaborators.

MarqRazz (approved):
Tested locally with my own package and this fixes my testing issues.

bmagyar (approved):
Much needed, thank you!

sea-bass:
@christophfroehlich could you give me a green approve oh behalf of Marq and Bence, so I can merge this and cut the release? 🙏🏻

pac48 (approved):

christophfroehlich (approved):

imcelroy:
> Oh, interesting. I had not seen that warning for some reason. That is... not ideal.
> 
> Yeah, it's a problem because this isn't a "regular" Python module where all the code is available in the source folder.
> 
> I guess if we go back to the state before this PR, the only bulletproof way to test these generared modules is to do it from another package.
> 
> Anyways, would suggest we keep iterating and if we find something good we can release again. Thanks!

@sea-bass I agree with iterating. I only wanted to point out that this will break existing unit tests for users that are trying to adapt `generate_parameter_library` into an existing ROS2 python package (like myself) after having followed ROS guidelines.

I think part of the issue is that the generated parameter module is treated as something "compiled" like in C++. However, from what I understand that isn't how Python works, which is why everything is in the source folder.

As a potential solution to the source control issue that you mentioned here: https://github.com/PickNikRobotics/generate_parameter_library/issues/233#issuecomment-2584885231, the user could add it to their `.gitignore`.

Thank you again for your help!

PR #245, opened 2025-02-18
Update format.yaml

mosfet80:
Ubuntu 22.04  support ended 4 months and 3 weeks ago (30 Sep 2024)


sea-bass:
I think Ubuntu 22.04 is LTS until 2027: https://ubuntu.com/about/release-cycle

christophfroehlich:
releveant is the [github runner] https://github.com/actions/runner-images), but only 20.04 is deprecated as of now.

Issue #6, opened 2022-07-21
Register async callback for updates


tylerjw:
Provide an interface for registering async callbacks for parameters being updated. Be careful that the mutex is not locked when these functions are called so if they call `get()` it doesn't create a deadlock.

```cpp
ParameterListener::update()
{
  {
    std::lock_guard lock(mutex_);
    // modify params_ variable
  }
  std::lock_guard lock(update_callbacks_);
  for (auto const& f : update_callbacks_) f(params_);
}

class ParameterListener {
 public:
  void register_callback(std::function<void(Params const&)> cb) {
    std::lock_guard lock(update_callbacks_);
    update_callbacks_.push_back(cb);
  }
 private:
  std::mutex callbacks_mutex_;
  std::vecotr<std::function<void(Params const&)>> update_callbacks_;
```

Warn users that the lifetime of the callback function has to be at least as long as the lifetime of the ParameterListener object.

PR #250, opened 2025-03-24
Feat: add user callback

YannickdeHoop:
Hi,

I added functionality to bind a custom callback. This callback will be fired when the parameters are updated, and those updated parameters are then directly available. I think this covers https://github.com/PickNikRobotics/generate_parameter_library/issues/6

### DIFF
--- a/example/src/minimal_publisher.cpp
+++ b/example/src/minimal_publisher.cpp
@@ -77,7 +77,7 @@ void MinimalPublisher::reconfigure_callback(
     const admittance_controller::Params& params) {
   RCLCPP_INFO(get_logger(), "Reconfigure callback fired!");
   RCLCPP_INFO(get_logger(), "New control frame parameter is: '%s'",
-            params.control.frame.id.c_str());
+              params.control.frame.id.c_str());
   RCLCPP_INFO(get_logger(), "fixed string is: '%s'",
               std::string{params.fixed_string}.c_str());
   const auto fixed_array = params.fixed_array;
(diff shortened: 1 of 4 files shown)


Timple (approved):

YannickdeHoop:
ping @pac48

example/src/minimal_publisher.cpp
```diff
@@ -71,4 +73,9 @@ void MinimalPublisher::timer_callback() {
   }
 }
 
+void MinimalPublisher::reconfigure_callback(
+    const admittance_controller::Params& params) {
+  RCLCPP_INFO(get_logger(), "Reconfigure callback fired!");
```
pac48, line 78: Can we put the name of the parameter in this logged message?

pac48 (commented):
Looks good, I just added a couple of comments. Is it possible to implement this callback for Python as well? If not, let's make an issue to track it. 

Edit: ah I see [this](https://github.com/PickNikRobotics/generate_parameter_library/pull/254) PR does the Python implementation

generate_parameter_library_py/generate_parameter_library_py/jinja_templates/cpp/parameter_library_header
```diff
@@ -182,6 +182,9 @@ struct StackParams {
 {%- endif %}
       updated_params.__stamp = clock_.now();
       update_internal_params(updated_params);
+      if (user_callback_) {
+         user_callback_(updated_params);
```
Timple, line 186: The variable is called `update_params`. But I believe it contains all parameters, not just the ones that got updated right?

Would it be an idea to only supply the actually updated parameters? Otherwise the logging statements (like in the example) would be really verbose if you have multiple parameters. And from a log-reader standpoint you still don't have a clue what actually changed.

Timple: Sure, can do!

One last question: is `setUserCallback` the right name here to differentiate these two use cases?
Because once this is in, it's harder to rename because of backward compatibility etc.

pac48:
@YannickdeHoop Can you run pre-commit to fix the CI failure

pac48 (approved):

PR #260, opened 2025-05-04
Improve `try_get_params` method

saikishor:
Only return true when the parameters are modified and return false for others cases
I've also modified it to use the unique_lock as it is much better than try_lock on the mutex itself

### DIFF
--- a/generate_parameter_library_py/generate_parameter_library_py/jinja_templates/cpp/parameter_library_header
+++ b/generate_parameter_library_py/generate_parameter_library_py/jinja_templates/cpp/parameter_library_header
@@ -125,12 +125,12 @@ struct StackParams {
     }
 
     bool try_get_params(Params & params_in) const {
-      if (mutex_.try_lock()) {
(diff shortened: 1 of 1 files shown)


christophfroehlich (approved):

sea-bass (commented):
Apologies for the delay -- I'm not really on these repos much these days.

Given this is a change in behavior, I don't feel like I should have the authority to make that decision. I'm going to leave this up to @pac48 as an original author of the package.

christophfroehlich:
This was introduced with #205, but as discussed in the linked PR I think that it should have a different behavior. We could also deprecated the current one and add the proposed method with a different name if you prefer.

pac48 (approved):

christophfroehlich (approved):
Great. Let us also add both methods to the README..

nbbrooks:
Thank you for the updates.

@christophfroehlich since we added the new method as opposed to changing the ABI, I assume we should get this PR into the next release? I am going to work on that in the morning, sorry it is taking so long.

PR #265, opened 2025-06-15
Update README.md

yassine-cherni:
The previous link didn’t point to the intended section, fixed now!
Just one step closer to becoming the next CTO ;")


christophfroehlich:
> The previous link didn’t point to the intended section, fixed now!

which link? You only changed a header?

yassine-cherni:
Yes, it just was not Point to the section related.

christophfroehlich:
?
image

yassine-cherni:
int
Just when we click here, it doesn't go to related section ;")

PR #268, opened 2025-07-13
fix setuptools deprecations

mosfet80:
FIX:

#UserWarning: Unknown distribution option: 'tests_require'

and

#SetuptoolsDeprecationWarning: License classifiers are deprecated. !!


Plumezz:
Yes, I also modified the `setup.py` like this. But I did not notice the License issue.

PR #271, opened 2025-07-24
fix: make dataclasss public

MrBlenny:
Currently dataclasses generated in python look like:
```
class admittance_controller:

    class Params:
        # for detecting if the parameter struct has been updated
        stamp_ = Time()
        ...
        class __Pid:
            rate = 0.005
            class __MapJoints:
                p = 1.0
                i = None
                d = 1.0
            __map_type = __MapJoints
            def add_entry(self, name):
                if not hasattr(self, name):
                    setattr(self, name, self.__map_type())
                return getattr(self, name)
            def get_entry(self, name):
                return getattr(self, name)
        pid = __Pid()
```

The problem with this is you cannot access the child classes using `admittance_controller.Params.__Pid` because of the __. This PR changes this so the classes are no longer private.

```
class admittance_controller:

    class Params:
        # for detecting if the parameter struct has been updated
        stamp_ = Time()
        ...
        class __Pid:
            rate = 0.005
            class MapJoints:
                p = 1.0
                i = None
                d = 1.0
            _map_type = MapJoints
            def add_entry(self, name):
                if not hasattr(self, name):
                    setattr(self, name, self.__map_type())
                return getattr(self, name)
            def get_entry(self, name):
                return getattr(self, name)
        pid = Pid()
```

Given they were private before I don't expect this will cause any backwards compatbility issues.

### DIFF
--- a/generate_parameter_library_py/generate_parameter_library_py/jinja_templates/python/declare_struct
+++ b/generate_parameter_library_py/generate_parameter_library_py/jinja_templates/python/declare_struct
@@ -1,4 +1,4 @@
-class __{{struct_name}}:
+class {{struct_name}}:
 {%- filter indent(width=4) %}
 {% if struct_fields|length -%}
 {{struct_fields-}}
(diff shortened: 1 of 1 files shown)


christophfroehlich (commented):
Why do you need this being public?

MrBlenny:
So children of the Params class can be used as type annotations and passed into functions.

saikishor:
> So children of the Params class can be used as type annotations and passed into functions.

I understand that you would like to use type annotations, but the GPL is not meant for that right?. IMO It is just a way to access the parameters, but not to construct classes

christophfroehlich (changes requested):
I would like to go forward with this PR.

Could you please

- merge the master into this branch (I can't do that because you opened the PR from an organization. alternatively, you could add me as committer to your orgs fork).
- Add a test as suggested by copilot to ensure this will work in future?

PR #272, opened 2025-07-25
fix setuptools deprecation

mosfet80:
FIX:

#UserWarning: Unknown distribution option: 'tests_require'


christophfroehlich (approved):
Thanks, a follow-up to #268

PR #275, opened 2025-09-02
fix(generate_parameter_library_py): nested mappings double dot bug (#10)

natrad100:
* fixes a bug with nested mapping of parameters in python (removes the double dot bug)

Minor bug where two dots appear where only a single dot should.

```python
param_name = f"{self.prefix_}nested_param.{value_1}..{value_2}.inner_param"
```

now

```python
param_name = f"{self.prefix_}nested_param.{value_1}.{value_2}.inner_param"
```

Let me know if you'd like me to re-write the test to align better with what you have (or remove)


christophfroehlich (changes requested):
I see the issue also in the current example_python.

But instead of writing a pytest for the exported code, what do you think of adding integration tests
https://github.com/PickNikRobotics/generate_parameter_library/issues/25?

christophfroehlich:
>     * Can you please merge the main branch into your fork's branch? (github does not allow maintainers to change branches from an organization; you also could add me as contributor to the respective repo)

Created a new PR #329 to fix pre-commit and test it with current main branch.

PR #280, opened 2025-10-27
Account for nested map params (#277)

### LINKED ISSUE #277: Incorrect codegen behaviour with nested structure and non-string values inside parameter map
If you use a nested structure within the parameter map, it will generate invalid code.

Example:
```yaml
example:
  keys:
    type: string_array

  __map_keys:
    string_value:
      type: string
    
    bool_value:
      type: bool

    nested:
      foo:
        type: int
      bar:
        type: string
      baz:
        type: bool
```

this will cause it to generate code like:
```cpp
StackParams get_stack_params() {
  Params params = get_params();
  StackParams output;
  output.__map_keys.nested.foo = params.__map_keys.nested.foo;
  output.__map_keys.nested.baz = params.__map_keys.nested.baz;

  return output;
}
```

notice how it *doesn't* have the `bar` parameter, but it *does* have the `foo` and `baz` parameters.

agonzat:
Fixes how mapped params are handled so that nested params within map fields are also accounted for when generating the update/declaration blocks.

E.g.:
```yaml
example:
  keys:
    type: string_array

  key_structs:
    __map_keys:
      worked_fine:
        type: int
      nested:
        didnt_work: # This
          type: int
```

Still if multiple maps are stacked they need to be contiguous (same as before).

Fixes #277 .

### DIFF
--- a/parameter_traits/include/parameter_traits/parameter_traits.hpp
+++ b/parameter_traits/include/parameter_traits/parameter_traits.hpp
@@ -43,8 +43,8 @@ template <typename... Args>
 [[deprecated(
     "When returning tl::expected<void, std::string> you can call fmt::format "
     "directly.")]] auto
-ERROR(const std::string& format,
-      Args... args) -> tl::expected<void, std::string> {
+ERROR(const std::string& format, Args... args)
+    -> tl::expected<void, std::string> {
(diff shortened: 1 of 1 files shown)


christophfroehlich (approved):
LGTM

diegoferigo-rai (approved):

PR #303, opened 2026-01-05
Use a temporary build directory for Python module output

bjsowa:
The `generate_parameter_module` macro outputs the Python module to `CMAKE_INSTALL_DIR` during the build phase. This results in build failures on the ROS buildfarm due to permission errors (example [here](https://build.ros2.org/job/Jbin_uN64__leo_example_line_follower__ubuntu_noble_amd64__binary/33/console))

This PR modifies the macro to output the module to a temporary directory in `CMAKE_CURRENT_BINARY_DIR` and install it to `CMAKE_INSTALL_DIR` during the install phase.

### DIFF
--- a/generate_parameter_library/cmake/generate_parameter_library.cmake
+++ b/generate_parameter_library/cmake/generate_parameter_library.cmake
@@ -117,14 +117,17 @@ function(generate_parameter_module LIB_NAME YAML_FILE)
   # Set the yaml file parameter to be relative to the current source dir
   set(YAML_FILE ${CMAKE_CURRENT_SOURCE_DIR}/${YAML_FILE})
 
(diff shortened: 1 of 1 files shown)


christophfroehlich:
If the current setup fails on the buildfarm, why haven't that failed earlier? Or is it that simple, nobody released a python node using this library?

bjsowa:
@christophfroehlich Any chance this gets merged and released soon? I'm getting a lot of spam from ROS buildfarm about failing jobs...

christophfroehlich:
Oh my bad, I have tested the options with the wrong package above :see_no_evil: Just tested now also successfully with `cmake_generate_parameter_module_example`. 
I'll dig into the build farm error.

christophfroehlich:
I hoped that we could fix the failing tests on the buildfarm before, but nobody could help so far.. I'll trigger a release tonight.

PR #306, opened 2026-01-16
Fix nested mapped parameters map

### LINKED ISSUE #278: No way to reference non top-level keys in parameters map
There is no way to reference a non top-level key in a parameter map

Here is an example:
```yaml
example:
  nested:
    entries:
      type: string_array

    __map_entries:
      value:
        type: string
```

this does not properly find the `entries` key, and instead generates code that looks like the following:
```cpp
// error: no member named 'entries' in 'example::Params'
for (const auto & value_1 : updated_params.entries) {
//                          ~~~~~~~~~~~~~~ ^

    /* ... */
}
```

the following also does not work:
```yaml
example:
  nested:
    keys:
      type: string_array

    __map_nested.keys:
      value:
        type: string
```
as it will generate code that looks like the following:
```cpp
struct Params {
    struct Nested {
        std::vector<std::string> keys;
// error: expected member name or ';' after declaration specifiers
        struct MapNested.Keys {
//      ~~~~~~~~~~~~~~~~^
            std::string value;
        };
// error: expected '>'
// error: expected member name or ';' after declaration specifiers
        std::map<std::string, MapNested.Keys> nested_keys_map;
//                                     ^
    } nested;
    // for detecting if the parameter struct has been updated
    rclcpp::Time __stamp;
};

/* ... */

// error: invalid range expression of type 'struct Nested'; no viable 'begin' function available
for (const auto & value_1 : updated_params.nested) {
//                        ^ ~~~~~~~~~~~~~~

// error: no member named '__map_nested' in 'example::Params::Nested'
    auto& entry = updated_params.nested.__map_nested.nested_map[value_1];
//                ~~~~~~~~~~~~~~~~~~~~~ ^

    /* ... */
}
```

NickLaurenson-Visionick:
Hello,  this PR aim to fix #278. 

As mention in the issue, the root caused came from `get_dynamic_mapped_parameter_names`, not return the full nested path:
if the the field to be mapped was `.nested_field.final_field`, the function return `final_field` causing error in the generated code.

Regarding the test, I wrote some basic one, let me know if need to do something else

### DIFF
--- a/example_python/generate_parameter_module_example/minimal_publisher.py
+++ b/example_python/generate_parameter_module_example/minimal_publisher.py
@@ -55,11 +55,11 @@ class MinimalParam(rclpy.node.Node):
 
         self.get_logger().info(
             "self.params.nested_map.entry1.value = '%s'"
-            % self.params.nested_map.entry1.value
+            % self.params.nested_map.get_entry('entry1').value
         )
         self.get_logger().info(
             "self.params.nested_map.entry2.value = '%s'"
-            % self.params.nested_map.entry2.value
+            % self.params.nested_map.get_entry('entry2').value
         )
 
     def timer_callback(self):


christophfroehlich (approved):
Thanks for the iterations. I tried it out on my own now, and added some test output to the examples.

PR #336, opened 2026-03-30
Fix conflicting validators

### LINKED ISSUE #243: Exception when using lt<> and gt<> together / exclusive bounds<>?
Since bounds<> is inclusive I tried using lt<> and gt<> together to achieve an exclusive bounds check, but I get the following error:
```
[node-1] terminate called after throwing an instance of 'std::length_error'
[node-1]   what():  Exceeded upper bound
```
Is this something worth fixing or are there plans to add an exclusive bounds check?

christophfroehlich:
Combination of lower and upper bounds resulted in invalid ROS code 
```c++
          descriptor.description = "specifies maximum acceleration limits for x, y and z axis";
          descriptor.read_only = false;
          descriptor.floating_point_range.resize(1);
          descriptor.floating_point_range.at(0).from_value = -10.0;
          descriptor.floating_point_range.at(0).to_value = std::numeric_limits<double>::max();
          descriptor.floating_point_range.resize(2);
          descriptor.floating_point_range.at(1).from_value = std::numeric_limits<double>::lowest();
          descriptor.floating_point_range.at(1).to_value = 10.0;
```
where floating_point_range must not be of size greater 1.
It throws `std::length_error("Exceeded upper bound")` (but bounds of this vector size, not of the declared parameter) resulting in
```
1: unknown file: Failure
1: C++ exception with description "Exceeded upper bound" thrown in SetUp().
```

~~This PR adds build-time errors for this unsupported combination. I don't see a point in auto-merging it, users of this lib are simply encouraged to just use the correct validators.~~
Edit: This PR adds build-time errors for unsupported combinations of two-sided bounds with one-sided bounds, but merges given one-sided bounds.

Fixes #334


christophfroehlich:
:eyes: @vincent-daniele

vincent-daniele:
Indeed, I see no point to use both because of the element_bounds validator. But I guess it should be warned in the doc that both upper and lower bounds cannot be used at the same time

christophfroehlich:
good point..ln the meantime I would see a point if we would support inclusive and exclusive bounds (gt+lt_eq).

christophfroehlich:
@vincent-daniele I added a simple note to the README, may I ask for another review?

christophfroehlich:
After consideration of #243 I implemented an auto-merge of one-sided bounds. Only the mixture of two-sided with one-sided bounds results in a compile time-error.

vincent-daniele (approved):

PR #337, opened 2026-03-30
Fix markdown generation for bounds/element-bounds

### LINKED ISSUE #253: Markdown validation conversion for bounds and element bounds wrong
When using `bounds<>` (e.g. `bounds<>: [0, 10]`) or `element_bounds<>` (e.g. `element_bounds<>: [0.0, 1.0]`) they are wrongly outputted to Markdown using `generate_parameter_library_markdown`.

In the outputted markdown it will simply put (for `bounds`)
```
parameter must be within bounds 0
```
and (for `element_bounds`):
```
each element of array must be within bounds 0.0
```

If instead, these are put in double square brackets, the Markdown output is wrong but the CPP conversion fails...

I would guess that this comes from using `arguments[0]` in [here](https://github.com/PickNikRobotics/generate_parameter_library/blob/main/generate_parameter_library_py/generate_parameter_library_py/generate_markdown.py#L80-L83)
```
        if validation.__contains__('VALUES'):
            validation = validation.replace('VALUES', str(arguments[0]))
        elif arguments:
            validation += ': ' + str(arguments[0])
```

christophfroehlich:
Resolves #253 

Now 

```
generate_parameter_library_markdown --input_yaml example/src/parameters.yaml --output_markdown_file parameters.md
```
prints in markdown and similarly in rst language

```markdown
## one_number
 - parameter must be within bounds [1024, 65535]
```

```markdown
## admittance.mass
*Constraints:*
 - length must be equal to 6
 - each element of array must be within bounds [0.0001, 1000000.0]
```

### DIFF
--- a/generate_parameter_library_py/generate_parameter_library_py/generate_markdown.py
+++ b/generate_parameter_library_py/generate_parameter_library_py/generate_markdown.py
@@ -67,6 +67,14 @@ class ParameterValidationMarkdown:
 
         self.validation = validation
 
+    @staticmethod
+    def format_arguments(arguments):
+        if not arguments:
+            return ''
+        if len(arguments) == 1:
+            return str(arguments[0])
+        return str(arguments)
+
     def get_validation_type(self, function_base_name):
         if function_base_name in self.sentence_conventions:
             return self.sentence_conventions[function_base_name]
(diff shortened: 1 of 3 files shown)


christophfroehlich:
:eyes: @JulianKu

christophfroehlich:
@JulianKu could you please test this PR and leave a review here? (everyone can do a review)

bmagyar (commented):
Do we want tests for this?

christophfroehlich:
> Do we want tests for this?

yes. The only thing is that this string comparison might be error prone (at least we had an issue with such a test and fmt generated output on different platforms).

destogl (approved):

PR #338, opened 2026-04-07
Fix rst syntax

Mat198:
This PR aims to solve #326. It changes how Python reads the string to 'raw mode', so it doesn't use backslash as an escape character. 

Unfortunately, it causes an error if the string ends in a backslash, as the interpreter can't see the closing string character (`'` or `"`). I think it's not worth addressing this, as the user can add a white space if a backlash at the end of the description is really necessary.

Here is the expected error if someone Google it in the future:

```bash
print(r'testing raw\')
  File "<stdin>", line 1
    print(r'testing raw\') # Just remove the \ at the end
          ^
SyntaxError: unterminated string literal (detected at line 1)
```

### DIFF
--- a/example_python/generate_parameter_module_example/minimal_publisher.py
+++ b/example_python/generate_parameter_module_example/minimal_publisher.py
@@ -81,6 +81,7 @@ def main(args=None):
         rclpy.spin(node)
     except KeyboardInterrupt:
         pass
-    
+
+
 if __name__ == '__main__':
     main()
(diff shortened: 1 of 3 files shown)


christophfroehlich (changes requested):
Thank you for the fix, I have successfully tested it.

Please just fix pre-commit (and maybe install it for the future)

Mat198:
Sorry for the precommit error. I didn't know I could run it locally :sweat_smile:

christophfroehlich:
Not only run it locally, but even automatically if you install it as a git pre-commit hook ;) (`pre-commit install`)

christophfroehlich (approved):
Thank you!

Mat198:
I'll check it out! Thank you!

PR #339, opened 2026-04-07
Fix ParameterDescription of exclusive bounds

### LINKED ISSUE #256: gt<> validator sets incorrect from_value in parameter descriptor
The ranges in the parameter descriptor are inclusive (https://github.com/ros2/rcl_interfaces/blob/rolling/rcl_interfaces/msg/FloatingPointRange.msg#L3), but currently the `gt<>` validator sets the lower limit to the specified value (instead of the next larger value):

https://github.com/PickNikRobotics/generate_parameter_library/blob/13a1043416b62393d284f067c0c52d4db547a175/generate_parameter_library_py/generate_parameter_library_py/jinja_templates/cpp/declare_parameter#L17

This results in inconsistent behavior between parameter descriptor and parameter validation:

```console
jonas@ade $ ros2 param describe /example_node background.gt_zero_float                            
Parameter name: background.gt_zero_float
  Type: double
  Description: Parameter with default value 1.0 and gt<>: 0.0 validation
  Constraints:
    Min value: 0.0
    Max value: 1.7976931348623157e+308
jonas@ade $ ros2 param set /example_node background.gt_zero_float 0.0 
Setting parameter failed: Parameter 'background.gt_zero_float' with the value '0' must be greater than '0'
```

christophfroehlich:
Fixes #256


ottojo (approved):
That looks like it will fix the issue i reported! (Haven't built and tested it so far, myself i can do that next week).
I remember when looking into it before i came across [`std::nextafter`](https://en.cppreference.com/cpp/numeric/math/nextafter), but i'm not sure if that is any different, the +/- epsilon is definetly more readable 👍

christophfroehlich:
Oh, `std::nextafter` seems to be a nice method. I haven't thought too much about floating point precision. Is `anynumber+eps` always greater than the floating point representation of `anynumber`? maybe not

christophfroehlich:
@ottojo have you had a chance to give that PR a try?

christophfroehlich:
:bell: @ottojo

ottojo (approved):
Sorry for the long delay! I just built your branch and tested a few values with the `gt<>`/`lt<>`/`gt_eq<>` validators, and everything seems to be correct. This also in particular fixes the bug reported in #256, thanks!

PR #344, opened 2026-04-12
Warn on conflicting validators

christophfroehlich:
Similar to  #336, but print a warning instead of throwing an exception 
```
$ colcon build --packages-select generate_parameter_module_example 
Starting >>> generate_parameter_module_example
--- stderr: generate_parameter_module_example                   

WARNING: Parameter 'conflicting_bounds' cannot combine 'bounds' with scalar bound validators (gt/gt_eq/lt/lt_eq). Use only 'bounds<>' for inclusive ranges, or only scalar bound validators.
---
Finished <<< generate_parameter_module_example [3.75s]

Summary: 1 package finished [4.01s]
  1 package had stderr output: generate_parameter_module_example
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -264,6 +264,8 @@ The built-in validator functions provided by this package are:
 | gt_eq<>  | [value]             | parameter >= value                   |
 | one_of<> | [[val1, val2, ...]] | Value is one of the specified values |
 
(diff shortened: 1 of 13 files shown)


christophfroehlich:
@vincent-daniele this is a suggestion in backporting this to the non-rolling distros, what do you think?

vincent-daniele:
Nice idea, would help avoid unclear errors

PR #345, opened 2026-04-12
Generate markdown documentation also for none type

### LINKED ISSUE #266: Documentation not being generated for parameters with type = none
Hello!

From my understanding of this part of the readme:

> In some cases, parameters might be unknown only at compile-time, and cannot be part of the generated C++ code. However, for documentation purpose of such parameters, the type none was introduced.

Running this command:

```
generate_parameter_library_markdown --input_yaml force_torque_sensor_broadcaster_parameters.yaml --output_markdown test.md
```

Over this parameter yaml:

```yaml
force_torque_broadcaster_controller:
  sensor_name:
    type: string
    default_value: ""
    description: "Name of the sensor used as prefix for interfaces if there are no individual interface names defined."
  frame_id:
    type: string
    default_value: ""
    description: "Sensor's frame_id in which values are published."
  sensor_filter_chain:
    type: none
    description: "Map of parameters that defines a filter chain, containing filterN as key and underlying map of parameters needed for a specific filter. See <some docs> for more details."
```

Should yield a markdown with the documentation with the `sensor_filter_chain` param included. However, this doesn't seem to be the case, especially considering the implementation [here](https://github.com/PickNikRobotics/generate_parameter_library/blob/ba34f27c9eccfc9d55aa47edce909ae0a1305df9/generate_parameter_library_py/generate_parameter_library_py/generate_markdown.py#L263C5-L263C46). 

I would just like to know if this is indeed the intended behavior. As far as my understanding goes, the only documentation for "none" parameters will be in the YAML itself. I guess this is not especially important as the parameters are normally defined elsewhere in the plugins. 

Thanks a lot!

christophfroehlich:
Closes #266

```bash
$ generate_parameter_library_markdown --input_yaml src/ros2_controllers/chained_filter_controller/src/chained_filter_parameters.yaml --output_markdown test.md
```

Before:

> # Chained Filter Parameters
> 
> Default Config
> ```yaml
> chained_filter:
>   ros__parameters:
>     input_interfaces: '{}'
>     output_interfaces: '{}'
> 
> ```
> 
> ## input_interfaces
> 
> Name of the input state interface
> 
> * Type: `string_array`
> * Default Value: {}
> * Read only: True
> 
> *Constraints:*
>  - parameter is not empty
> 
> *Additional Constraints:*
> 
> 
> 
> ## output_interfaces
> 
> Name of the output state interface
> 
> * Type: `string_array`
> * Default Value: {}
> * Read only: True
> 
> *Constraints:*
>  - parameter is not empty
> 
> *Additional Constraints:*

After:

> # Chained Filter Parameters
> 
> Default Config
> ```yaml
> chained_filter:
>   ros__parameters:
>     <input_interfaces>:
>       filter_chain: ''
>     filter_chain: ''
>     input_interfaces: '{}'
>     output_interfaces: '{}'
> 
> ```
> 
> ## input_interfaces
> 
> Name of the input state interface
> 
> 
> * Type: `string_array`
> 
> * Default Value: {}
> * Read only: True
> 
> *Constraints:*
>  - parameter is not empty
> 
> *Additional Constraints:*
> 
> 
> 
> ## output_interfaces
> 
> Name of the output state interface
> 
> 
> * Type: `string_array`
> 
> * Default Value: {}
> * Read only: True
> 
> *Constraints:*
>  - parameter is not empty
> 
> *Additional Constraints:*
> 
> 
> 
> ## filter_chain
> 
> Map of parameters that defines a filter chain, containing filterN as key. Valid for all interfaces, overrides <input_interfaces>.filter_chain if set. The fields for each filter are: type: The filter plugin to be loaded name: Actual name semantically describing the filter, e.g., low_pass_filter params: And underlying map of parameters needed for a specific filter, refer to the specific filter documentation.
> 
> 
> 
> ## <input_interfaces>.filter_chain
> 
> Instead of a single configuration, a distinct configuration can be provided for each input interface. The key is the name of the input interface, and the value is a map of parameters that defines a filter chain, containing filterN as key. The fields for each filter are: type: The filter plugin to be loaded name: Actual name semantically describing the filter, e.g., low_pass_filter params: And underlying map of parameters needed for a specific filter, refer to the specific filter

### DIFF
--- a/generate_parameter_library_py/generate_parameter_library_py/jinja_templates/markdown/parameter_detail
+++ b/generate_parameter_library_py/generate_parameter_library_py/jinja_templates/markdown/parameter_detail
@@ -2,7 +2,9 @@
 {% if description|length %}
 {{description}}
 {% endif %}
+{% if type != 'none' %}
 * Type: `{{type}}`
+{% endif %}
 {%- if default_value|length %}
 * Default Value: {{default_value}}{% endif %}{% if read_only %}
 * Read only: {{read_only}}{% endif %}{%- if constraints|length %}
--- a/generate_parameter_library_py/generate_parameter_library_py/jinja_templates/rst/parameter_detail
+++ b/generate_parameter_library_py/generate_parameter_library_py/jinja_templates/rst/parameter_detail
@@ -1,4 +1,4 @@
-{{name}} ({{type}}){%- filter indent(width=2) %}{% if description|length %}
+{{name}}{% if type != 'none' %} ({{type}}){% endif %}{%- filter indent(width=2) %}{% if description|length %}
 {{description}}
 {% endif %}
 {%- if read_only %}
--- a/generate_parameter_library_py/generate_parameter_library_py/parse_yaml.py
+++ b/generate_parameter_library_py/generate_parameter_library_py/parse_yaml.py
@@ -855,8 +855,12 @@ class GenerateCode:
             validations,
             additional_constraints,
         ) = preprocess_inputs(self.language, name, value, nested_name_list)
-        # skip accepted params that do not generate code
-        if code_gen_variable.lang_type is None:
+        # Skip accepted params that do not generate code for code-generation targets.
+        # Documentation targets still need to keep these entries (e.g. type: none).
+        if code_gen_variable.lang_type is None and self.language not in {
+            'markdown',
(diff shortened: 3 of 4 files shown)


bmagyar (approved):
thank you!

PR #348, opened 2026-04-17
Fix install paths

christophfroehlich:
**This was completely coded by GPT-5.3-Codex and needs proper review and testing.**

## Summary

Fix Python generated-parameter module installation/import behavior across all colcon layouts (standard, isolated, and merge-install), resolving intermittent prerelease import failures.

## Problem

Tests could fail with:

- `ModuleNotFoundError: No module named 'generate_parameter_module_example.admittance_parameters'`
- `ModuleNotFoundError: No module named 'generate_parameter_module_example.custom_validation'`

Root causes were path/layout mismatches in generation output handling and empty package `__init__.py` files in install space.

## Changes

- Updated [src/generate_parameter_library/generate_parameter_library_py/generate_parameter_library_py/setup_helper.py](src/generate_parameter_library/generate_parameter_library_py/generate_parameter_library_py/setup_helper.py):
  - Parse setuptools args robustly (`--opt value` and `--opt=value`)
  - Support additional install/build flags (`--build-lib`, `--install-lib`, `--prefix`, `--root`)
  - Correctly map `build*` roots to matching `install*` roots (including `build_isolated` -> `install_isolated`)
  - Auto-detect merge vs isolated layout from `AMENT_PREFIX_PATH`
  - Avoid generating into source-resolving symlink paths

- Updated [src/generate_parameter_library/generate_parameter_library_py/generate_parameter_library_py/generate_python_module.py](src/generate_parameter_library/generate_parameter_library_py/generate_parameter_library_py/generate_python_module.py):
  - Populate generated package `__init__.py` with namespace path extension when missing or empty

- Added [src/generate_parameter_library/example_python/generate_parameter_module_example/__init__.py](src/generate_parameter_library/example_python/generate_parameter_module_example/__init__.py):
  - Namespace path extension for split source/install package resolution

## Validation

All scenarios pass for `generate_parameter_module_example`:

- Standard layout: `48 passed, 0 failed`
- Isolated layout (`--build-base/--install-base`): `48 passed, 0 failed`
- Merge-install layout (`--merge-install`): `48 passed, 0 failed`

Generated module is now written to the correct install location in each layout.

PR #356, opened 2026-05-28
add gitignore for generated files in example_python

ottojo:
When having the repo checked out locally, building the workspace leads to untracked files in this git repo. I think it's safe to just gitignore them since they are autogenerated anyways.

### DIFF
--- a/example_python/generate_parameter_module_example/.gitignore
+++ b/example_python/generate_parameter_module_example/.gitignore
@@ -1,0 +1,2 @@
+__init__.py
+admittance_parameters.py


christophfroehlich:
this is only a problem with --symlink-install?

ottojo:
Yes, i can confirm this is only an issue with symlink installs

christophfroehlich:
there are several issues with that, which I tried to solve with #348 but lacks of reviews from other people. Maybe you could also try it and give me feedback.
in the meantime it's ok for me to merge this

christophfroehlich (approved):

PR #357, opened 2026-06-03
Accept quotes in description and additional constraints

### LINKED ISSUE #353: Accept quotes in descriptions
Currently, a quote (`"`) in the description breaks the C++ and Python syntax.
In C++, it should be an easy fix with raw string literals. In Python, we simply could add something like `.replace('"', '\\"')` in valid_string_python.

Hugal31:
Transform strings to replace `"` with `\"` in C++ and Python.

I tried the raw-string route (e.g.

```
R"(my "quoted" description
with line-breaks)";
```

but it didn't work nicely with the jinja indent filter.

Closes #353


christophfroehlich (changes requested):
Thank you! The changes LGTM, output in markdown and rst also looks fine.

But
- please fix pre-commit
- Could you leave a review for sanity checking in https://github.com/PickNikRobotics/generate_parameter_library/pull/361 please? I'd like to symlink it instead of changing three files here.

christophfroehlich (approved):
Thank you, LGTM!

PR #358, opened 2026-06-09
[CI] Add compatibility CI build

christophfroehlich:
This tests the rolling version on stable distros (kilted, jazzy, humble as of now). This ensures that users can use the latest (breaking) changes by building this package from source in their workspace.


christophfroehlich:
Failing test on humble is known

PR #360, opened 2026-06-09
Fix floating_point_range numerical values

christophfroehlich:
Fixes error on humble like
```
- generate_parameter_module_example.test.test_params_consistency.TestParamsConsistency test_use_feedforward_commanded_input
  <<< failure message
    AssertionError: The 'to_value' field must be of type 'float'
  >>>

```

From Iron on, the assertions are disabled per default: https://docs.ros.org/en/rolling/Releases/Release-Iron-Irwini.html#optional-argument-that-hides-assertions-for-messages-class

Question is: Should we instead harden the behavior and wrap the values in a float()?

### DIFF
--- a/example/src/parameters.yaml
+++ b/example/src/parameters.yaml
@@ -286,30 +286,30 @@ admittance_controller:
       lt<>: [ 20 ]
   lt_eq_fifteen_double:
     type: double
-    default_value: 1
+    default_value: 1.0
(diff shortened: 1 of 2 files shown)


saikishor (approved):
Looks fine to me.

In ROS 1, the params used to auto cast automatically, I don't know if we should do the same here, but I'm okay with this change

PR #361, opened 2026-06-09
Use a shared symlinked parameter file in all examples

christophfroehlich:
so that we won't introduce any issues from one language to another

Copy and translate the test_params_consistency test to c++ (by Claude Haiku 4.5)


Hugal31:
I would have made `example_python` install the .yaml file and `example` and `example_cmake_python` depends on `example_python`, but this works.

I also tested it against Bloom and it resolves the symlink before creating the separate branch per package.

PR #365, opened 2026-06-26
Improve Python custom validator example

aryanputta:
## Summary

- updates the Python example custom validator module so it demonstrates a real validation failure path
- wires the example YAML to the named damping-ratio validator
- documents how Python examples pass a custom validation module through `generate_parameter_module`

Fixes #321.

## Validation

- `python3 -m py_compile example_python/generate_parameter_module_example/custom_validation.py`
- direct validator behavior checks for valid, wrong-size, non-finite, non-positive, and out-of-bounds inputs
- generated the Python module from `example_python/generate_parameter_module_example/parameters.yaml` and parsed it with `ast.parse`
- `git diff --check`
- `env SKIP=ament_copyright pre-commit run --files README.md example_python/generate_parameter_module_example/custom_validation.py example_python/generate_parameter_module_example/parameters.yaml`

`ament_copyright` is not installed in my local environment, so I could not run that hook locally. All other touched-file pre-commit hooks passed.


christophfroehlich (changes requested):
Thank you for improving this.

Please use the same logic as in the c++ example, or adapt both. I try to have the same content in both examples. Can you also leave a review in #361 for that reason please?

PR #366, opened 2026-07-07
Create SECURITY.md for security policy

tiffanydawnatkins-cloud:
Add a security policy document outlining supported versions and vulnerability reporting.
2660d537d8a82796b034879b2884a81d3af0c44c


christophfroehlich (changes requested):
I don't think that this is needed/useful in it's current state here.

PR #378, opened 2026-08-09
replace lowest/max double bounds with +/- infinity

### LINKED ISSUE #376: `.inf` not within floating point range
Since the last lyrical update, the following:
```YAML
admittance_controller:
  my_param:
    type: double
    default_value: .inf
    validation:
      gt_eq<>: 0
```
causes:
```
  what():  parameter 'my_param' could not be set: Parameter {my_param} doesn't comply with floating point range.
```

This can be reproduced with the `generate_parameter_library_example`, adding the `my_param` parameter above and running:
```
ros2 run generate_parameter_library_example test_node --ros-args --params-file src/generate_parameter_library/example/config/implementation.yaml
```

I cannot pinpoint the exact change that causes this. It's not just caused by the 1.2 to 1.3 update; there must be other package updates involved. In any case, the above example used to work and I think it should continue to work

christianrauch:
The double bounds use `lowest`/`max` on C++, while the Python implementation uses +/- `float('inf')`.

Align the implementations by replacing `lowest`/`max` with +/- `infinity`.

Original issue discovered by @pum1k in https://github.com/PickNikRobotics/generate_parameter_library/issues/376#issuecomment-5231462166.

Fixes #376 .


example_python/generate_parameter_module_example/parameters.yaml
```diff
@@ -345,6 +345,11 @@ admittance_controller:
     type: int_array_fixed_5
     default_value: [3,3,3]
     read_only: true
+  double_no_upper_bound:
+    type: double
+    default_value: .inf
+    validation:
+      gt_eq<>: 0
```
christianrauch, line 352: Note to myself: The change appears in `example_python` because the C++ parameter file is symlinked to this one.

christophfroehlich:
kind of duplicate of #169. Have you checked if rqt_reconfigure handles this properly? I remember opening a PR there, but not sure now

christianrauch:
> kind of duplicate of #169. Have you checked if rqt_reconfigure handles this properly? I remember opening a PR there, but not sure now

Do you remember what exactly the problem with `rqt_reconfigure` was?

When I start the `rqt_reconfigure` plugin via
```
QT_QPA_PLATFORM=xcb rqt --standalone rqt_reconfigure.param_plugin.ParamPlugin
```
and set a value, it is applied and I can retrieve it again via:
```sh
ros2 param get /admittance_controller double_no_upper_bound
```

The slider goes from 0 to inf, so using the slider does not make too much sense. But setting it manually via the text field works as expected.

I get a couple of "The range of this parameter is too large for the slider to handle. Scaling down to fit within 32 bits with factor 2.328306435454494e-10." warnings, but those apply to some integer values.

christophfroehlich:
This is fine then. As far as I remember, rqt_reconfigure stopped listing parameters of this node if any inf limits were configured.

christophfroehlich (approved):

christophfroehlich (changes requested):
On the humble compatibility build we see that we have assertions in the python code because infinity is not within the float range
https://github.com/PickNikRobotics/generate_parameter_library/pull/360

should we just disable the checks of the message?

PR #383, opened 2026-08-28
fix that if no parameter is updated, the user callback is not called

fabianhirmann:
This can be the case if one node has multiple param listeners in multiple namespaces and then the `add_on_set_parameters_callback` is called for all the parameter changes and not just for the relevant ones.

### DIFF
--- a/example/test/example_test_gtest.cpp
+++ b/example/test/example_test_gtest.cpp
@@ -85,7 +85,8 @@ TEST_F(ExampleTest, try_get_params) {
 TEST_F(ExampleTest, update_unrelated_param) {
   ASSERT_TRUE(param_listener_->try_get_params(params_));
 
-  // if some unrelated parameter is updated, the params_ should not be updated/considered as old
+  // if some unrelated parameter is updated, the params_ should not be
+  // updated/considered as old
   example_test_node_->declare_parameter("some_other_parameter", 42);
   ASSERT_FALSE(param_listener_->is_old(params_));
   const rclcpp ::Parameter new_param("some_other_parameter", 43);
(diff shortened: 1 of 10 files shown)


fabianhirmann:
After the pull request is merged, please backport the change to at least ROS 2 Humble as well.

christophfroehlich (changes requested):
Description sounds reasonable to be fixed, but can you please add tests for that?

fabianhirmann:
> Description sounds reasonable to be fixed, but can you please add tests for that?

Thank you for your quick response! I also added now a test in `example/test/example_test_gtest.cpp` because as far as I saw at the other test cases there is no testing of the behavior of the produced `ParamListener`. The newly created test passes with my changes and fails when tested without my changes.

By that I also noticed that I missed to update the generator for `update_runtime_parameter` but this is now also fixed (all tests should pass now (at least this is the case when testing the changes on ROS Humble)).

christophfroehlich (commented):
Thanks for the test, can you add one for the python module as well please?

fabianhirmann:
> Thanks for the test, can you add one for the python module as well please?

Thank you for the catch! I also added now a test for the python module.

fabianhirmann:
Hi @christophfroehlich!

I fixed the linter errors and general tests and as far as I see all _regular_ tests are passing. However the windows build is failing but the fail seems to be unrelated to my PR. Could you support in that? Or is it currently to be expected that the windows build is failing?

Many thanks,
Fabian

christophfroehlich (approved):
I fixed the win build, now lgtm

PR #384, opened 2026-09-01
tl_expected is deprecated

mosfet80:
info: https://index.ros.org/p/tl_expected/


christophfroehlich:
No. `tl-expected` is provided from the system dependency libexpected-dev
https://github.com/PickNikRobotics/generate_parameter_library/blob/02e8f0d2d4ab897934e993969ca04b9681405864/generate_parameter_library/package.xml#L26

that's why we deprecated  `tl_expected`

PR #388, opened 2026-09-17
Add a standalone configuration validator (#312)

### LINKED ISSUE #312: [Feature Request] Standalone validator tool
## Summary

Add a CLI tool (e.g. `generate_parameter_library_validate`) that validates a ROS 2 parameter config YAML against one or more parameter definition YAMLs, without requiring a ROS 2 runtime, a colcon build, or a running node.

## Motivation

Currently, the only way to validate whether a config YAML conforms to the parameter definitions is to build the generated code, launch a node with `--params-file`, and observe whether it throws. This works during development, but creates pain in two important scenarios:

**CI pipelines:** A robotics project may have dozens of parameter definition YAMLs across packages and a large monolithic (or composed) config YAML that is deployed to the robot. When someone changes a parameter value in that config, the only feedback loop today is a full colcon build + node launch. A lightweight validation step in CI (similar to a linter) would catch issues like typos in parameter names, wrong types, or out-of-bounds values in seconds rather than minutes.

**Field engineers and integrators:** People who configure robots on-site often do not have a full build environment. They edit YAML config files and deploy them. A wrong parameter name or an out-of-range value currently only surfaces as a runtime crash or cryptic error message on the robot. A standalone validation tool that they can run on their laptop (with just a `pip install`) would save significant debugging time and prevent deploying broken configs to production robots.

The library already has `generate_parameter_library_markdown` as a precedent for a standalone tool that reads the parameter definition YAML and produces useful output without requiring ROS 2. A validator would follow the same pattern.

## Proposed interface

```bash
# Validate a single config against a single parameter definition
generate_parameter_library_validate \
  --param-definition src/my_package/config/my_parameters.yaml \
  --config config/robot.yaml

# Validate against multiple parameter definitions (common for composed systems)
generate_parameter_library_validate \
  --param-definition src/pkg_a/config/a_parameters.yaml \
  --param-definition src/pkg_b/config/b_parameters.yaml \
  --config config/robot.yaml

# Strict mode: also warn about parameters in the config
# that do not appear in any definition (catches typos)
generate_parameter_library_validate \
  --param-definition src/my_package/config/my_parameters.yaml \
  --config config/robot.yaml \
  --strict
```

Exit code 0 on success, non-zero on validation failure, with human-readable error messages:

```
ERROR: my_node.background.r: value 300 violates bounds<>[0, 255]
ERROR: my_node.background.colour: unknown parameter (did you mean 'color'?)
ERROR: my_node.pid.rate: expected type 'double', got 'string' ("fast")
WARNING: my_node.background.b: missing from config, will use default_value 0
```

## Scope of validation

The parameter definition YAML format already encodes rich type and constraint information. The following built-in validators all map cleanly to static checks that can be performed without a ROS 2 runtime:

**Value validators:** `bounds<>`, `lt<>`, `gt<>`, `lt_eq<>`, `gt_eq<>`, `one_of<>`

**String validators:** `fixed_size<>`, `size_gt<>`, `size_lt<>`, `not_empty<>`, `one_of<>`

**Array validators:** `unique<>`, `subset_of<>`, `fixed_size<>`, `size_gt<>`, `size_lt<>`, `not_empty<>`, `element_bounds<>`, `lower_element_bounds<>`, `upper_element_bounds<>`

The tool should also check:
- **Type correctness** (e.g. int vs double vs string vs arrays)
- **Required parameters** (parameters without `default_value` must be present in the config)
- **Namespace structure** (the `ros__parameters` nesting in config YAMLs)
- **Mapped parameters** (`__map_<key>` where possible, i.e. when the key parameter has a `default_value` or is present in the config)

**Explicitly out of scope (with warning):**
- **Custom C++ validator functions** (e.g. `"my_project::integer_equal_value": [3]`). These are arbitrary code and cannot be evaluated offline. The tool should emit a warning: `INFO: my_node.param: custom validator 'my_project::integer_equal_value' cannot be checked offline, skipped`
- **`read_only`** is a runtime behavioral constraint, not checkable offline

## Implementation considerations

- The tool should be a pure Python CLI with no dependency on ROS 2 packages, so it can be installed via `pip` in any environment. This matches how `generate_parameter_library_markdown` works.
- The existing YAML parsing logic in `generate_parameter_library_py.parse_yaml` already understands the full parameter definition DSL and could be reused as the foundation.
- The config YAML parser needs to handle the `ros__parameters` namespace and the `/**` or node-specific namespacing conventions.
- Consider also supporting a `--output-json-schema` flag that converts the parameter definitions to JSON Schema. This would enable integration with existing JSON Schema tooling and IDE support (e.g. YAML language server in VS Code for autocompletion and inline validation of config files).

## Example CI usage

```yaml
# GitHub Actions example
- name: Validate robot config
  run: |
    pip install generate_parameter_library
    generate_parameter_library_validate \
      --param-definition src/navigation/config/nav_parameters.yaml \
      --param-definition src/behavior/config/behavior_parameters.yaml \
      --config deploy/config/robot_production.yaml \
      --strict
```

## Related

- `generate_parameter_library_markdown` already demonstrates the pattern of a standalone CLI tool that processes parameter definition YAMLs
- Issue #59 reports confusing runtime error messages for invalid configs, which offline validation would help prevent
- Issue #30 discusses separating validation logic, which aligns with making validation available outside of generated code

facontidavide:
@tylerjw hello !

I hope you don't mind too much the AI generated PR.
I found myself needing this and I would like to contribute upstream, so that more people may benefit from it.

let me know what you think

# Description

Add generate_parameter_library_validate, a CLI that checks a ROS 2 parameter configuration file against one or more parameter definition files. It needs no ROS 2 installation, no colcon workspace and no running node, so it can run as a CI step or on the laptop of somebody configuring a robot.

The tool reuses the existing parser and the validators in python_validators.py, so the rules it applies are the ones the generated code applies at runtime. It reports a value whose type does not match the declared type, a value rejected by a built-in validator, and a parameter declared without a default_value that the configuration does not set. --strict also reports parameters that no definition declares, with a suggestion when the name is close to a declared one, and parameters absent from the configuration that will take their default.

Custom validator functions are C++ and cannot run outside a build, so they are reported as skipped. read_only describes runtime behaviour and is out of scope.

CodeGenVariableBase now keeps the size of a fixed size type on the instance. It was computed and discarded, and the validator needs it to check array lengths.


tylerjw:
@facontidavide I don't mind at all. I'll have my bob review this and will try it out myself. I don't have permissions to merge things here anymore but I'm happy to review stuff.

generate_parameter_library_py/generate_parameter_library_py/validate_config.py
```diff
@@ -0,0 +1,652 @@
...
+    actual = yaml_type_name(value)
+    if actual == expected:
+        return None
+    if expected == 'double' and actual == 'int':
+        return (
+            f"expected type 'double', got 'int' ({value}); ROS 2 does not convert "
+            f'integers to doubles, write {value}.0'
+        )
+    if reads_as_number(value, expected):
+        return None
```
tylerjw, line 155: [P1] Preserve quoting when resolving numeric types

A configuration containing `rate: "2.0"` passes this check for a parameter declared as `double`: `reads_as_number()` accepts the string and `coerce_value()` converts it to a float. I reproduced the CLI returning 0. ROS 2 preserves quoted scalars as strings, so the node rejects that override instead. This also affects quoted integers and array elements. Please retain scalar style/tag information while loading YAML and resolve the ROS type independently of the expected declaration, so unquoted scientific notation remains supported without accepting quoted numbers.

Reference: https://github.com/ros2/rcl/blob/rolling/rcl_yaml_param_parser/src/parse.c#L227-L248

generate_parameter_library_py/generate_parameter_library_py/validate_config.py
```diff
@@ -0,0 +1,652 @@
...
+            elif strict:
+                diagnostics.append(
+                    Diagnostic(
+                        WARNING,
+                        where,
+                        f'missing from config, will use default_value '
+                        f'{parameter.default_value}',
+                    )
+                )
+            continue
```
tylerjw, line 478: [P2] Validate the effective default when a parameter is omitted

This `continue` skips all validation for omitted parameters that have a default. For example, a `string_array` declared with `default_value: []` and `validation: {not_empty<>: null}` passes the CLI with `my_node: {ros__parameters: {}}` (exit 0), although the generated runtime initialization validates the effective value and rejects the empty array. A numeric default outside its declared bounds has the same problem. Please run the applicable validators on the effective default before continuing; the strict-mode warning alone does not detect the startup failure.

generate_parameter_library_py/generate_parameter_library_py/validate_config.py
```diff
@@ -0,0 +1,652 @@
...
+    sections = {}
+    for key, value in document.items():
+        if not isinstance(value, dict):
+            continue
+        if ROS_PARAMETERS_KEY in value:
+            sections[str(key)] = value[ROS_PARAMETERS_KEY] or {}
+            continue
+        for nested_key, nested_value in value.items():
+            if isinstance(nested_value, dict) and ROS_PARAMETERS_KEY in nested_value:
+                sections[f'{key}/{nested_key}'] = nested_value[ROS_PARAMETERS_KEY] or {}
```
tylerjw, line 430: [P2] Discover node sections recursively through namespaces

Only the first two mapping levels are searched for `ros__parameters`. A valid ROS parameter file nested as `robot -> arm -> my_node -> ros__parameters` is therefore treated as an unnamed plain parameter tree. I reproduced `rate: -5.0` under that section returning exit 0 with a definition that has `default_value: 1.0` and `gt<>: [0.0]`; the actual override is never checked. In strict mode it instead produces unknown-parameter errors for the flattened namespace. Please recurse through namespace mappings until reaching each `ros__parameters` section.

ROS's parser accumulates namespace components until it reaches that key: https://github.com/ros2/rcl/blob/rolling/rcl_yaml_param_parser/src/parse.c#L902-L938

tylerjw (commented):
Hi @facontidavide — I'm Bob, Tyler's AI assistant, posting this review with Tyler's explicit permission.

I reproduced three cases where the validator exits successfully despite a configuration that fails at runtime; the inline comments include details. The validator and mapped-parameter tests pass (38 tests). I couldn't collect the full Python suite because ament_index_python is unavailable in this environment.

christophfroehlich (changes requested):
Thanks for taking this one!

Should we directly use this in this repo for integration testing the existing configuration files?

@nitschi has originally proposed this, could you have a look please and leave a review here?

